Issue #183, opened 2019-01-06
Reading HidGuardian registry data crashes options menu


WhiteMagic:
HidGuardian writes not only VID&PID to the registry for devices but also a serial number if the device provides one via the USB protocol. This is not expected by Gremlin as most devices don't, however, some do and as a result the regular expression parsing the value fails to separate the string properly resulting in a crash when attempting to parse the returned values as integers.

https://cdn.discordapp.com/attachments/528501166385135636/531564131246145547/unknown.png

WhiteMagic:
Thanks to jmriego for providing the fix to this issue.

PR #184, opened 2019-01-10
Fix crash when reading HiDGuardian registry

jmriego:
This pull request solves the issue #183 .
I am not capturing the serial number as it seems like it's not necessary, but if it was, this code could do that:
`split_regex = re.compile("HID\\\\VID_([a-zA-Z0-9]+)&PID_([a-zA-Z0-9]+)(?:&REV_([a-zA-Z0-9]+))?")`

Honored to do the first pull request! I really like your app after trying every other similar one like UCR, Freepie, Glovepie... and I have several years of experience with Python. So I would be happy to help with some other issues if you want

### DIFF
--- a/gremlin/hid_guardian.py
+++ b/gremlin/hid_guardian.py
@@ -224,14 +224,21 @@ class HidGuardian:
 
         # Process each entry to extract vendor and product id
         device_data = []
-        split_regex = re.compile("HID\\\\VID_([a-zA-Z0-9]+)&PID_([a-zA-Z0-9]+)")
+        split_regex = re.compile("HID\\\\VID_(.{4})&PID_(.{4})")
         for entry in data[0]:
             match = split_regex.match(entry)
             if match:
-                device_data.append((
-                    int(match.group(1), 16),
-                    int(match.group(2), 16)
-                ))
+                try:
+                    device_data.append((
+                        int(match.group(1), 16),
+                        int(match.group(2), 16)
+                    ))
+                except ValueError:
+                    gremlin.util.display_error(
+                        "Failed to extract vendor and product id for HidGuardian entry:\n\n{}"
+                            .format(entry)
+                    )
+
 
         return device_data
 
--- a/gremlin/hid_guardian.py
+++ b/gremlin/hid_guardian.py
@@ -224,7 +224,7 @@ class HidGuardian:
 
         # Process each entry to extract vendor and product id
         device_data = []
-        split_regex = re.compile("HID\\\\VID_(.+)&PID_(.+)")
+        split_regex = re.compile("HID\\\\VID_([a-zA-Z0-9]+)&PID_([a-zA-Z0-9]+)")
         for entry in data[0]:
             match = split_regex.match(entry)
             if match:


WhiteMagic:
Thanks for the fix, though I think this would not solve the issue entirely, as the regex is used with match method which requires to find the complete expression as opposed to just a part of it. From looking at the cases where the code failed it is caused by an additional term existing on top of the VID and PID the current code is looking for.

Capturing the REV value probably is also not going to work as I have one example that contains another third term (https://dto9r5vaiz7bu.cloudfront.net/v012w5fh7xvnw/tavern_upload_large.png) so it seems that VID and PID are guaranteed to exist with additional random stuff.

I think the safest solution is to have regex similar to this:
```re.compile("^HID\\\\VID_([a-fA-F0-9]{4})&PID_([a-fA-F0-9]{4}).*$")```
This should capture 2 groups of 4 hex values with the potential of additional stuff at the end. 

As these two matches are then later on interpreted as hex values and the failing of that cast causes the problem it probably can't hurt to put that part into a try catch and log + display in case of a failure and then continue so Gremlin doesn't crash but the user is informed of the issue.

**Comment by jmriego** [comment]
Happy to help! I think both ways should work. The match method returns a match if the regex is found at the start of the string. Trying both options, they seem to give the same result here:

```
def test_regex(entry, split_regex):
    split_regex = re.compile(split_regex)
    match = split_regex.match(entry)
    return [match_group(match, i) for i in range(4)]

print test_regex('HID\VID_046D&PID_C216&REV_0300', "HID\\\\VID_([a-zA-Z0-9]+)&PID_([a-zA-Z0-9]+)")
print test_regex('HID\VID_046D&PID_C216', "HID\\\\VID_([a-zA-Z0-9]+)&PID_([a-zA-Z0-9]+)")
print test_regex('HID\VID_046D&PID_C216', "^HID\\\\VID_([a-fA-F0-9]{4})&PID_([a-fA-F0-9]{4}).*$")
print test_regex('HID\VID_046D&PID_C216', "^HID\\\\VID_([a-fA-F0-9]{4})&PID_([a-fA-F0-9]{4}).*$")
```

Anyway, I'll have a look at the logging functions JoystickGremlin is using and modify the code. Thanks for checking this!

jmriego:
I have modified the code to add an error message when not being able to load from the HidGuardian registry. I have modified the regex to get the four characters next to PID and VID even if they are not an hexadecimal number. That way, if there's a problem, it will show a dialog instead of not hitting the re.match part and masking that there is a registry entry it cannot read.
I can change it back if necessary. By the way, I'd rebase all the commits together before the merge to keep a cleaner history

WhiteMagic:
Thanks for this I'll have a look over it tomorrow but from the quick glance it looks like this should work fine.

WhiteMagic:
Thanks for your work on fixing this. At the moment I'm doing some kind of open heart surgery with replacing SDL2 so there might be quite a bit of underlying change. Things that tend to be somewhat independent are containers and actions.

jmriego:
No worries! Always good to give back a bit of help after using such a nice program. Good luck! I wouldn't even know where to start looking into SDL2 or alternatives 😆 
I'll have a look at the open issues and see what can I do related to containers and actions. Cheers

Issue #145, opened 2018-09-18
Smart toggle container


WhiteMagic:
Add a new container implementing a smart toggle activation type. This container would simulate "holding down a key" if a short press is detected, releasing on the second press. If a longer press is detected they action is released as soon as the physical key is released.

WhiteMagic:
Implemented by [jmriego](/jmriego).

PR #186, opened 2019-01-22
Smart toggle container

jmriego:
Hi! I'm adding some code here for having a smart toggle container as mentioned in #145 .

The code is copied from the basic container and adding as little changes as possible for that.
Not sure if that's the correct way or should I reuse the code in some way as I believe the tempo container does?

### DIFF
--- a/joystick_gremlin.spec
+++ b/joystick_gremlin.spec
@@ -22,14 +22,13 @@ for root, _, files in os.walk("container_plugins"):
 added_files = [
     ("about", "about"),
     ("doc", "doc"),
-    ("gfx", "gfx"),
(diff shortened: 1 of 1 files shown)


WhiteMagic:
Thanks! I'll look over it in a bit, but as far as I remember all containers and actions are independent of each other (at least in their code) as the code that loads them into Gremlin technically is intended to be able to load them from arbitrary locations, such that someone could write their own set of custom actions and containers without them needing to be part of the official Gremlin installer. Though that feature isn't exposed in a configurable way yet.

jmriego:
That sounds good! Learning little by little :)

PR #188, opened 2019-02-03
Smart toggle container

jmriego:
Hi!

I have made some changes after what we talked and I believe the code is almost ready to be merged.
This code includes two types of toggle: hold and momentary.
The idea is that in some games, some functions could be toggled on and off every time you press the button, and in others the function would be on only while the button is active.

It also includes on press and on release options similarly to the tempo container.

I'm not sure about stopping the map action from doing an auto-release. I'm just hacking it with this code, but there's probably a cleaner way.
`self.action_set.functors[1].needs_auto_release = False`

I hope the code is clear enough. Just ping me if I can improve anything. Hope this does the trick!

### DIFF
--- a/container_plugins/smart_toggle/__init__.py
+++ b/container_plugins/smart_toggle/__init__.py
@@ -1,0 +1,369 @@
+# -*- coding: utf-8; -*-
+
+# Copyright (C) 2015 - 2019 Lionel Ott
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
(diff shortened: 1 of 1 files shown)


WhiteMagic:
Thanks for your work, I'll have a look over it today or tomorrow.

jmriego:
Sure, no problem at all! Still learning about this but I'll try to help with some other containers/actions

WhiteMagic:
Finally got some time to look over the container, mostly I think it does what we want it to do. Some things I think we should change though.

- The toggle type shouldn't be necessary as that behavior can be achieved via normal actions and conditions. Maybe not as simply but the very different behaviour makes this confusing.
- The default acvitation type probably should be "on press" as that has the behaviour you'd expect from the feature, mainly the hold / toggle activating the action immediately. Having the other option is good but that's something for special use cases I'd imagine.

jmriego:
I tried to have all options I could think of even if things as the "on release" would probably not be needed usually. That way, I can just remove whatever doesn't make sense to include in JG.
The reason why I added the toggle type is because in Elite Dangerous you can have some functions mapped as a toggle or a hold type mapping and wanted to make this work with any option as I thought most games would use either one or the other.
Just for my own information if you don't mind... how would you do that with normal actions and conditions?

Anyway, no problem, I'll have the change ready for tomorrow.

jmriego:
Everything should be fine now. If needed, I can rebase the code to make it only one commit

WhiteMagic:
Thanks for the nice work. No need to squash the commit or similar, Github actually gives me the option to squash them into a single commit before merging if I wanted to.

Issue #144, opened 2018-09-18
Double tap container


WhiteMagic:
Add a new container type which requires a double tap to activate the contents.

muchimi:
While the tempo feature adds the ability to do a short press and a long press, another, although probably less common input change mode, is to distinguish between an action on single tap vs double tap. 

This would be another option below the tempo feature.

Use-cases:  (1) some avionics control systems require the pilot to pickle a button twice to either cycle or lock a target, (2) to select a waypoint on a map with a moving cursor.   (3) communication ack usually is a click-click (or double tap on the PTT key).

This can be done in JG currently with a custom module - it would be nice to have that built-in as another option below the tempo() function.

WhiteMagic:
Thanks to @jmriego for implementing this feature.

PR #190, opened 2019-02-16
Double Tap Container

jmriego:
Hi!

I have pushed some code here that should do the trick for the double tap container #144 . It has two different actions, one for single, and one for double tap.
It also has an option that, honestly, I didn't know how to name. If you go for a double tap, it can either run the first tap as soon as you press it, or wait until it's sure it was a double or a single tap before sending either event.

If there's any change I should be doing here, just let me know. No rush.

Thanks!

### DIFF
--- a/container_plugins/double_tap/__init__.py
+++ b/container_plugins/double_tap/__init__.py
@@ -346,7 +346,7 @@ class DoubleTapContainer(gremlin.base_classes.AbstractContainer):
 
         :return True if the container is configured properly, False otherwise
         """
-        return len(self.action_sets) == 2 and None not in self.action_sets
+        return any(len(action_set) for action_set in self.action_sets)
 
 
 # Plugin definitions
--- a/container_plugins/double_tap/__init__.py
+++ b/container_plugins/double_tap/__init__.py
@@ -1,0 +1,355 @@
+# -*- coding: utf-8; -*-
+
+# Copyright (C) 2015 - 2019 Lionel Ott
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import copy
+import logging
+import threading
+import time
+from xml.etree import ElementTree
+
+from PyQt5 import QtWidgets
+
+import gremlin
+import gremlin.ui.common
+import gremlin.ui.input_item
+
+
+class DoubleTapContainerWidget(gremlin.ui.input_item.AbstractContainerWidget):
+
+    """DoubleTap container for actions for double or single taps."""
+
+    def __init__(self, profile_data, parent=None):
+        """Creates a new instance.
+
+        :param profile_data the profile data represented by this widget
+        :param parent the parent of this widget
+        """
+        super().__init__(profile_data, parent)
+
+    def _create_action_ui(self):
+        """Creates the UI components."""
+        self.profile_data.create_or_delete_virtual_button()
+
+        self.options_layout = QtWidgets.QHBoxLayout()
+
+        # Activation delay
+        self.options_layout.addWidget(
+            QtWidgets.QLabel("<b>Double-tap delay: </b>")
+        )
+        self.delay_input = gremlin.ui.common.DynamicDoubleSpinBox()
+        self.delay_input.setRange(0.1, 2.0)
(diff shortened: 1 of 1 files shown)


WhiteMagic:
I'll take a proper look at this once I get a bit of time. I haven't thought much about what the container should contain / what it should do.

Your current setup allows there to be a different action for single vs. double tap which allows having different actions just as with the tempo container. Another option would be to have a double tap only option, though I think that's less desirable given it would not allow to trigger either the single or the double tap action. That sounds better to me than the other option.

jmriego:
Sounds good! Yes, I thought it could be possible to need to execute either the single or the double but not both at the same time. That's mainly why I didn't create a double-tap only container with only one action.
But anyway, you would know better what we can create with the conditions, etc.
Thanks!

WhiteMagic:
Finally got some time to look over this. Overall I'd say this does what would be expected of the container. Some changes / issues that I ran into. The comments below are really just minor things and nothing broken about the container itself which works just fine as is already.

1. The `_is_container_valid` will always return true the way things are setup. A better check would be that at least one of the action sets contains an action. I moved away from having the requirement of a container having all possible cases populated (tempo and hat to buttons come to mind). Essentially, if parts of a container are used a container is supposed to be considered valid and only if it cannot possibly ever do anything is it invalid.
2. If you wanted to get rid of the `self.value_press` and `self.event_press` values you could do what I did in the tempo container with the lambda expression, binding all values into the `_single_tap` method which would the only require storing the `self.value_release` and `self.event_release` variables.
3. I'd use spaces in "Double Tap" whenever it's used for display purposes as the only reason. For example `DoubleTapContainer.name` can be "Double Tap" as this isn't used as an identifier anywhere and is mainly for human readability purposes.
4. I'd default to having the container wait for the double tap timeout before sending a single tap.

The description of when / whether to send single tap is a bit weird, while it's quite obvious what it will do maybe we can find a better or more succinct description for it. For the current "instantly" a possible alternative might be "always". For the waiting one "delayed" might be an option. Though that only covers the time aspect of it, when the other big difference is that "instantly" will trigger both while the "wait" one will only trigger one or the other. Based on that we could also use "combined" and "exclusive" to describe the activation mode overall?

As a general idea, which also goes for the tempo and chain containers is that the whole logic mess of what state is the container in could probably be done much nice using the very simplistic final state machine I have in fsm.py. Just randomly occurred to me while going over this, not really a need but probably something I'll take a look at some point.

jmriego:
Sure, I can make those changes. I'll have a better look but it shouldn't take me too long to do that.
The only thing I'm not completely sure about is the `_is_container_valid` thing. They way I found to find out if an action set contains an action is by making sure the `functors` attribute is not empty. Would that work?

WhiteMagic:
The functor thing only has a meaning when a profile is executed, however, the `_is_container_valid` is also used when saving a profile. Therefore, I typically look at the content of `action_sets` variable which will contain the action sets and actions associated with them as the UI essentially is just modifying those entries.

The hat buttons container https://github.com/WhiteMagic/JoystickGremlin/blob/features/container_plugins/hat_buttons/__init__.py here should give you a good idea of what should also work for the double tap container. It essentially just counts the number of action sets present and as long as at least one is present all is good. This obviously means that the functor needs to be able to handle all but one possible action set being empty, though I think constructing a execution graph object from an empty action set should just return a "no op" graph. So realistically that should not pose any problems.

jmriego:
Sorry for the delay, busy week!

I have made those changes you mentioned with the exception of the finite state machine. I didn't spend enough time understanding how to use that script but can look into it.
All the changes I have added are basically what you mentioned but removing the self.event_release variable was a bit more tricky.
Imagine that we have created a profile with a double tap with a delay = 0.5 and it's of type exclusive.
Let's suppose these two different cases:
- You long press the button for more than 0.5 seconds. The first 0.5 seconds nothing will happen but then at 0.5 the script knows it can't be a double click so it will send a release event (with a timer), then as soon as you release the button, the release event is processed.
- You press and release a single tap quickly. When you press it, the timer is created and will be run in 0.5 seconds. If you release it after 0.1 seconds, that leaves you 0.4 seconds until you know it's not a double tap and then you will send a short press and release. This is working by creating a 0.5 seconds timer for a press, and when you release it, you cancel that timer and create a different one sending a short press and release after 0.4 seconds.

I hope I explained well enough how this is working. Let me know if there's any concern

WhiteMagic:
Awesome, also no need to be sorry about "delays" it's awesome that you're taking your free time to add to Gremlin. I went ahead an merged the changes into Gremlin.

Yeah the whole FSM thing was anyways just something that occurred to me while looking over the code but definitely wasn't something I expected you to change, given I'm not even sure how much easier it would make things really. Your explanation is clear and yeah I figured there might be a bit more to the timers since the double / single combinations seemed like they might get messy and confusing to reason about all possible cases :-)

muchimi:
@jmriego Thanks for making this - very helpful and also thanks to @WhiteMagic for a wonderful framework.

jmriego:
sure thing! to be honest these things I'm helping you to add to JG I will also be using them myself.
So why not help when you made such an useful program?
It's also very well structured and easy to follow. So you can count on me adding a few more things :smiley:

Issue #242, opened 2019-09-13
Suggestion: Attempt to whitelist self with HID Guardian during boot


Tahvohck:
At the moment vJoy + Joystick Gremlin + HID Guardian is pretty much the best combination I could ask for, but one thing really bothers me currently, and it's that I have to separately whitelist Gremlin after launching it. If Gremlin could attempt to whitelist itself during boot, even without admin privileges, that would make it so I only ever have to launch one program (Gremlin) before launching a game. I know that this can be done via the v1 API of HG:

`http://localhost:26762/api/v1/hidguardian/whitelist/add/<PID>`
`http://localhost:26762/api/v1/hidguardian/whitelist/remove/<PID>`

I'd submit a pull request, but I'm not sure where in the program flow would be best. The code would be pretty simple:

```python
import requests, os
pidadd = "http://localhost:26762/api/v1/hidguardian/whitelist/add/{}"
try: requests.get( pidadd.format(os.getpid()) )
except: pass
```

If this was done before device polling, it would also mean that the devices would be immediately available.

WhiteMagic:
There hasn't been much additional work with regards to interfacing with HidGuardian, mainly because I was waiting for a new more refined version to be released. However, it seems that for the foreseeable future there won't be anything happening on that front.

The issue with using HidCerberus is that this works under the assumption that this has been installed alongside HidGuardian, which may not be the case. An additional problem is that Gremlin actually does a bit more then just whitelisting it's own PID. Gremlin annotates the PIDs that it added such that it can remove those upon launch to get rid of old PIDs that no longer are active. This would not be possible (at least directly) by using the HidCerberus route.

With regards to devices being available, this should already be the case, HidGuardian related actions happen before device listing etc. happens. If devices don't show up upon launch when HidGuardian is used with Gremlin launching as admin then there's something else wrong.

So overall I'm not sure if it's worth replacing the existing system for one that requires another system service to run. It would be nice not having to run Gremlin as root, however, I also don't know how many people have HidGuardian installed without HidCerberus.

Tahvohck:
That's part of why this can be done in a try block. The program can confirm that the whitelist has one of its old PIDs via the v1 API call of:
`http://localhost:26762/api/v1/hidguardian/whitelist/get`
[...]

PR #244, opened 2019-09-18
fix for #242 - Initial framework for HID Cerberus integration

Tahvohck:
Throwing this up as a draft as I've got the main things mentioned in #242 already coded in. It's not complete but the framework is there.

pycodestyle config will most likely be removed later - I committed it out of habit and never removed it.


Tahvohck:
I had to force-push this because it had some work-related email information that I needed to delete. As far as I can see, this successfully reset the author information on all commits without losing date info, basic history, etc.

Tahvohck:
This is now debugged in a running environment, and has the most basic of code to whitelist joystick gremlin during launch. HIDGuardian direct registry interfacing is not affected.

WhiteMagic:
Hey, that looks like quite a bit of work you've done there. I'll have to find some time to go over it look through it but already a big thanks for that.

shimasteama:
Hello
I joined github in order to get in contact with the developer of joystìck Gremlin.
I tested joystìck Gremlin and have a suggestion for the case when you have Two identical devices.
Does this Email reach the Developer?
Thank you for a reply

Harry
-- 
Diese Nachricht wurde von meinem Android Mobiltelefon mit 1&1 Mail gesendet.Am 19.09.19, 10:25, WhiteMagic <notifications@github.com> schrieb:
Hey, that looks like quite a bit of work you've done there. I'll have to find some time to go over it look through it but already a big thanks for that.

—You are receiving this because you are subscribed to this thread.Reply to this email directly, view it on GitHub, or mute the thread.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/WhiteMagic/JoystickGremlin/pull/244?email_source=notifications\u0026email_token=ALR67UMAJHBENQLEPEEPGELQKMZPXA5CNFSM4IXXRGA2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD7CVAAA#issuecomment-533024768",
"url": "https://github.com/WhiteMagic/JoystickGremlin/pull/244?email_source=notifications\u0026email_token=ALR67UMAJHBENQLEPEEPGELQKMZPXA5CNFSM4IXXRGA2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD7CVAAA#issuecomment-533024768",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

WhiteMagic:
> Hello I joined github in order to get in contact with the developer of joystìck Gremlin. I tested joystìck Gremlin and have a suggestion for the case when you have Two identical devices. Does this Email reach the Developer? Thank you for a reply Harry

Hey, if the suggestions is with regards to the pull request feel free to reply here, otherwise it's better to just open a new issue here https://github.com/WhiteMagic/JoystickGremlin/issues for the suggestion.

Tahvohck:
I should add that I've coded this to emulate the `hid_guardian.HidGaurdian` class, but I'm not sure that this needs to be a class at all. Why don't we break this out into a single `hid_cerberus` module before pulling the branch into WhiteMagic:develop ?

WhiteMagic:
I did some skimming over the code and it looks good. Though the thing I think that needs to happen is that both interfaces to hidguardian, i.e. registry mangling HidGuardian and the webinterface HidCerberus can be used interchangeably and transparently.

Currently they would be two different and independent pieces of code, furthermore it wouldn't even be possible to use the UI to whitelist / blacklist devices via the HidCerberus class even though it provides some function calls for it, as there is nothing hooked up to the UI side of things.

What I currently imagine as being the most sensible solution, especially with all the UI rewrite going on is for `hid_guardian.HidGuardian` to become the management class for HidGuardian related stuff which automatically selects a suitable backend, i.e. `Registry` or `HidCerberus` based on what is usable, if anything. The management class `HidGuardian` would then simply call into one of these backends to whitelist gremlin, add and remove devices, get device list etc. That way the existing UI can continue to work, no possible conflicts when both registry and hidcerberus are usable etc.

I probably also would stick all of that stuff inside the hid_guardian module as that simply arranges everything related to it, i.e. the management class and the two backends together into one file rather then being split across a bunch of different ones.

Tahvohck:
Sounds like a good way to do it. I think that the best switch would be whether or not the program was launched with admin rights, since if it was it can directly manage the Registry and not care if the REST API is running. Merging functionality shouldn't be too hard either once that switch is determined. From what I saw of the UI code you would also have to change how that particular UI panel disables itself, which can be a call to `hid_guardian.HidGaurdian` as well.

Tahvohck:
I've avoided adding anything to the registry provider so far since I'm not yet familiar with its code. However, the stubs are now in place, as is the Cerberus provider. Note that I made all of the methods classmethods, since there's no reason to instantiate separate provider instances each time (nothing in any of them need to be stateful)

WhiteMagic:
I'll look over it tonight and see if I can give some feedback. The registry based interface shouldn't be really complicated or need much changing since if we can just use that as the API reference and the HidCerberus interface can mimick it well enough we can just model based off the registry based one without needing to change much or anything.

Tahvohck:
I've actually already written up an interface class, now it's just a matter of making sure that the registry provider responds as expected to it. I was planning on doing some level of statefulness for the interface class actually, things like what devices are hidden already and whether or not the program has whitelisted itself.

WhiteMagic:
I don't really think adding statefulness is needed and likely will cause some grief. The state should be retrievable via the interface to HidGuardian. That way there is no need to make sure everything is kept in sync, as there is nothing preventing people whitelisting devices via means other then Gremlin. So my feeling is that adding this would not really give us anything we couldn't do already and would require a lot of careful synchronization and checking code.

Tahvohck:
Good point. All I added on that level so far was the backing data structures (AKA two empty lists), so that's simple enough to not add.

WhiteMagic:
I finally got some time to look over the changes. Overall I think it seems sensible though I probably would change the design a bit. The way it currently works is a bit opaque, due to the decorator class function lookup magic. The same behavior could be achieved using simple inheritance. Something along the following lines:

```python
import abs

class HidGuardianAbstract(metaclass=abc.ABCMeta:
    def __init__(self,):
        pass
    
    @abc.abstracmethod
    def add_device(self, identifier):
        pass

class HidGuardianRegistry(HidGuardianAbstract):
    def __init__(self):
        super().__init__()
        # Run initialization
    
    def add_device(self, identifier):
        # Registry implementation
        pass

class HidGuardianCerberus(HidGuardianAbstract):
    def __init__(self):
        super().__init__()
        # Run initialization
    
    def add_device(self, identifier):
        # HidCerberus implementation
        pass

def hg_interface():
    # Determine which interface to use
    has_registry = False
    has_cerberus = True
    if has_registry:
        return HidGuardianRegistry()
    elif has_cerberus:
        return HidGuardianCerberus()
    else:
        # Handle case of no interface
        return None
```
While this undoes the whole classmethod thing and technically there is no need to have methods with state, I think this makes it a bit clearer for anyone in the future looking over this. In Gremlin, there are only two places where the interface to HidGuardian is used. The main script which whitelists the instance of Gremlin and then the options UI dialog, which adds and removes devices from HidGuardian. As none of the places the interfaces are used in has high call rates, the overhead of creating or recreating the object isn't going to matter much.

Another minor note I try to adhere to PEP8 naming conventions, which is CamelCase for classes and names_width_underscores for variables and functions and methods.

Tahvohck:
I'll poke at this when I get a chance, Been kinda busy this past week. I'll remove the decorator class, which should make it far less opaque, and actually will keep the classmethod abstraction (it's really just a difference in how it's called). Edit: Actually, I'll look deeper into the abstract class thing. If this is a way that python has to enforce inheritance that's something I didn't realize was present in python.

As far as Pep8, I'll go in and fix all of that, I'm surprised my PEP8 linter (pycodestyle) didn't catch that.

WhiteMagic:
Yeah no worries take your time. And yeah the abc module allows Python to enforce abstract interfaces that need to be implemented by derived classes.

WhiteMagic:
HidCerberus and HidGuardian have been superseeded by HidHide and R14 won't contain direct UI support for the legacy tools anymore as using HidHide directly seems convenient enough.

PR #250, opened 2019-10-05
Added 'dejitter' funtion to remap

skombijohn:
I added a third "mode" to axis remaps:
Dejitter!

Dejitter will sample the last 5 event values in a "sliding window", averages them and compares this with the last saved average.
If the delta is big enough, the recently calculated average value will be saved as new average.
Eventually, the saved average will be forwarded to the vJoy axis.

**This removes jitter from physical joystick axis.**

There is still potential:

- allowing to control sampling amount (5 is hardcoded)
- allowing to control the delta threshold (0.005 is hardcoded)

This was just a quick shot at JoystickGremlin, which I discovered recently (yesterday) and found no solution to my jittery CH ProThrottle throttle axis...

And, it works beautifully!

I have no real python experience though, so I hope I did not f*** up stuff..

### DIFF
--- a/action_plugins/remap/__init__.py
+++ b/action_plugins/remap/__init__.py
@@ -94,10 +94,12 @@ class RemapWidget(gremlin.ui.input_item.AbstractActionWidget):
             self.absolute_checkbox = QtWidgets.QRadioButton("Absolute")
             self.absolute_checkbox.setChecked(True)
             self.relative_checkbox = QtWidgets.QRadioButton("Relative")
+            self.dejitter_checkbox = QtWidgets.QRadioButton("Dejitter")
             self.relative_scaling = gremlin.ui.common.DynamicDoubleSpinBox()
 
             self.remap_type_layout.addStretch()
             self.remap_type_layout.addWidget(self.absolute_checkbox)
+            self.remap_type_layout.addWidget(self.dejitter_checkbox)
             self.remap_type_layout.addWidget(self.relative_checkbox)
             self.remap_type_layout.addWidget(self.relative_scaling)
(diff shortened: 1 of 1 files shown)


WhiteMagic:
Hey, thanks for looking into this. I'm unlikely going to integrate a de-jitter function in this way, as it is not as generic as I would like it to be. It's a good stop-gap for your immediate issue I suspect but as a general feature, this has to be done in a different place in Gremlin, so that all other parts can benefit from it and not just a single action.

There are also some other functionalities that I would like to add with this sort of axis processing. Currently, Gremlin can do some axis calibration, for cases where the axis range is not the full possible one. Extending that functionality to add things such as noise filtering and axis rotation, for example, would be the way I'd implement this.

You can see what Gremlin is currently doing here https://github.com/WhiteMagic/JoystickGremlin/blob/develop/gremlin/event_handler.py#L217 where it simply applies the calibration to the axis input. This is what I'd extend into a more comprehensive axis filter that performs actions such as:
- apply calibration
- noise filtering
- axis rotation

Based on your filtering you only intend to handle cases where there is some small noise and no spikes in the signal. For that case the simple moving average filter you've implemented is adequate but to handle signals that have spikes some different filtering techniques would be used to remove those. This is where having all of this exposed as part of the calibration (probably would be renamed at that point) would be ideal as everyone will have different needs.

skombijohn:
Hey, thanks for your response. I fully understand and support your approach, so no worries 😉 

If I can be of help in the future, let me know.

Chris

WhiteMagic:
I've added an issue for the general problem and the system I think would be a decent approach of what your modification does for the remap action.

https://github.com/WhiteMagic/JoystickGremlin/issues/252

Issue #279, opened 2020-03-04
Axes centre when profile disables/enables


dixonte:
I very much like the feature to automatically load a profile based on the current application. However I've noticed when using it that if I tab out of my game for a moment and then back in, all the axes centre.

Concrete example. I have a HOSAS and 3 pedal setup, and I'm playing Star Citizen. I have the left stick's throttle set to the speed limiter and the right one's to the thrust limiter. The right and left pedals I have combined into a single access for forward/backward thrust. When I tab back into the game, both the speed and throttle limiters are set to 50%, and I have to stomp on the pedals to recalibrate them.

If I turn off the feature to load a profile based on the current application, this does not happen.

While ideally it would be nice to treat the root cause, I would like to also suggest an option that causes the profile to stay active until a different configured application becomes active.

Concrete example. Currently, if you have a profile set to load when cmd.exe is active, and a profile set to load when notepad.exe is active, then when explorer.exe becomes active Joystick Gremlin will completely deactivate. I would like an option such that if notepad was active then explorer becomes active, the notepad profile remains active until cmd becomes active, at which point the cmd profile is loaded and activated instead.

WhiteMagic:
This feature itself won't hurt, even if the underlying issue gets fixed, mainly because some games will still read the inputs even if they're in the background and you can end up sending unwanted inputs that way due to the vJoy devices being reset when Gremlin is disabled.

The change looks good, the only thing I would try to change is the text that goes with the checkbox, it's a bit long compared to other text on checkboxes I tend to use. So something more succinct would be ideal. What this effectively does is only load profiles and never deactivates them, so maybe that allows for a shorter text label. Maybe "Only swap profiles" "Keep profile active on focus loss" or similar ones? Setting the tooltip on the checkbox with an expanded explanation of what this does then would allow going into the detail and point of this feature.

The actual underlying issue is that Gremlin currently does not emit fake input events representing the physical state of devices on activation. Partially this is because Gremlin may not know the state of every input, due to the way DirectInput works, and so would have to guess for some. Ideally, this would be solved by adding hooks that run on things other than physical inputs, i.e. when gremlin is activated or deactivated, changing mode, pausing and unpausing etc. Though I've never gotten around to adding this system in. But as I mentioned this would not make the feature you propose superfluous, as deactivating still resets vJoy states so it may not be desired to deactivate the profile.

dixonte:
That sounds reasonable. I'll see about updating my pull request.

[...]

PR #281, opened 2020-03-07
Added feature: keep last automatically loaded profile active

dixonte:
Added a checkbox in the options that allows the user to choose to keep the last automatically loaded profile active when its application loses focus. This is something I need because disabling the profile causes my throttle and pedal axes to revert to centre when the profile disables, and I often tab out of Star Citizen to briefly chat on another monitor, etc.

I would consider this an adequate fix for the issue I raised, #279.


WhiteMagic:
Thanks for those improvements, finally got around to looking over them and merging them in.

Issue #271, opened 2020-01-11
Shift and Alt (and STRG?) Combos only work reliably as macros


BenBlake79:
I do not know if this occours everywhere, i had this issue with tempo and map to key for x-wing. Sometimes it works as it should - Shift + a gives an "A" - and sometimes it doesn't and gives an "a". So I used Macros with "press Shift - press a - release a - release Shift" to make it work. But this way it is not possible to hold the keys. I'd say there should be a short delay (maybe shorter than for macros is enough?) in "map to key" mode, maybe managable in the settings just in case that the time differs on different hardware in the same manner as it works for macro delay?

Great program, thank you very much. Just startet to use it as a replacement for TMs TARGET sw a week ago. How can I support you? Is it possible to make donations?

One thing that frustrates me is that it is not possible to change order of different actions or contaners without deleting the first and setting it or them up again. A drag and drop feature and copy/cut out and past feature (maybe with a second board as an extended clipboard) would be nice. Use it with the opened xml in Notepad++ at the moment and make such manipulations there an then load the profile again. For this workaround a simple reload button was helpful...

I am used to the Saitek Profiler and so setting up some features (Shift function as Joystick-Condition, axis splitting to make an input x times when moving an axis from -100 to +100 - my workaround was to make the calculations and code setup in Excel and pasting them in Notepad++) seem a little fiddly to me. An "autosplitter" which automaticly devides an axis in x parts and 1 or 2 or 3 shift-keys that can be set up globaly and then checked on the primary action page would be nice features. The same goes for the action condition which would be 2 checkboxes (pressed/released / fowards/backwards for axes) that can easily be fitted into the primary action page. This way it also would be easy to check much of a more complex setting with a single glance...

To define starting options for the drop down menus would be nice as well as it always starts with "map to button" and this is the action I (and most people?) use the least... (I think "map to key" is used the most btw.) 

I do not know if it is intended this way but when a vjoy device is set up as input device it is not longer possible to use it as a target so "physical joystick => vjoy#2 => vjoy#1" with manipulations on the way is not possible but until now no situation occured to me that could only be solved this way... When I thought I would need this feature I worked around somehow - I do not even know anymore what it was...

WhiteMagic:
[...]

PR #334, opened 2020-11-20
Map to Keyboard Tweak

amadeus:
Hello, I've been having a few issues using `Map to Keyboard` feature in IL2 due to what I think is caused by the release commands happening in the same order as they are pressed.

This creates issues when, for example binding flaps that depend on `F` to lower flaps and `LShift+F` to raise flaps. The Raise Flaps command ends up undoing itself due to the split second timing of `F` still being pressed when `LShift` is released.

Looking at the code, I was wondering if simply executing the release macro in reverse order would be a safe and quick way to fix this problem? (which as you can see is essentially what I did in this PR).

Unfortunately my Windows machine is not setup for development, so I was unable to test, and I also know jack shit about how Joystick Gremlin works under the hood, so there's a greater than 0 chance this won't work for other reasons that I may not fathom.

And if it is a valid fix, then the least I can do is help it along.

### DIFF
--- a/action_plugins/map_to_keyboard/__init__.py
+++ b/action_plugins/map_to_keyboard/__init__.py
@@ -105,7 +105,8 @@ class MapToKeyboardFunctor(AbstractFunctor):
             self.press.press(gremlin.macro.key_from_code(key[0], key[1]))
 
         self.release = gremlin.macro.Macro()
-        for key in action.keys:
+        # Execute release in reverse order
+        for key in reversed(action.keys):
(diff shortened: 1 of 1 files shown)


WhiteMagic:
That's been an issue for a while and the fix is that easy, similar to what was discussed in #271. I mainly hadn't fixed it because I'd end up touching on this when I get to rework all actions for R14, but your fix is what I would have implemented for it anyhow.

I'll merge into the develop branch even though that is massively behind R14, however, I will have to deal with consolidating the two at one stage anyhow.

PR #354, opened 2021-02-19
add Action Plugin - Virpil LED Control

dasTholo:
You can select the Devices, Command and the color with an Color Picker.
image

[quick test video](https://photos.app.goo.gl/jUC2BEy9v4Yf9ntM6)

Maybe I should create a file picker that takes the selection of "vpc_led_control.exe" to the virpil software suite.

Big Thanks to Whitemagic for this work! His code has inspired me


VegaAntares:
That is tip top for my alphas. Gonna merge that in my branch. Love the idea to turn the joystick red is case of heavy damages or yellow once hardpoints are deployed in Elite Dangerous by parsing the log journal !
Ideally this should be possible by plugin or addon mecanism for getting this feature in order to let the original software untied to some specific brands.
Edited: thanks for the good work there

WhiteMagic:
This looks really neat, two things I'd change:
1. I would probably put the led control logic into the plugin itself as I doubt any other part of Gremlin is going to use it.
2. I would also abuse the Configuration object (https://github.com/WhiteMagic/JoystickGremlin/blob/develop/gremlin/config.py) to store the path to the executable used to set the LED color. You can directly read/write the `_data` dictionary which isn't elegant but it will work. Just remember to call `save()` after writing so the data gets persisted. This way you can show the path in the plugin and the user can change it if needed but it will be saved by default.

The bigger issue is that while I can merge it, this code is likely never going to see a release. The reason is that I will only create a new release once R14 is ready. However, that has a different way in which the UI is done as it uses QML instead of raw Qt. So I would have to remove the plugin again as I won't be able to test it out when porting the UI component of it. The good part though is that the UI itself should be rather simple to create once R14 gets closer to being done and the logic of setting the LEDs would stay the same.

Unground33:
Thanks @WhiteMagic for JG and @dasTholo for this great extension. Really useful. Just started using it with the CM3 Throttle. Keen to support this going forward. Will be happy to test etc.

dasTholo (commented):
I´ve rewrite the Colorpicker to a Combolist, because the Virpil will only support a few Colors.

WhiteMagic:
As R14 will require a complete rewrite of all UI components this action won't be merged as there will be no R13 code-branch releases anymore. However, this PR still allows others to try this out.

TheGoodIdeaFairy:
I will definitely be giving this a try.  Much more elegant than the VPC_LEDs plugin I've been using... far less limiting to be able to assign a state change to a specific action rather than to a physical button from the plugin tab.  Thanks for this man!

WhiteMagic:
Would require re-implementation for R14, though then can be distributed independently of Gremlin as well due to user configurable plugin folder support.

Issue #358, opened 2021-03-04
FloatVariable ignores initial_value


henniss:
FloatVariable ignores it's initial_value argument if set to a float, because _init_numeric sets it to 0 unless the initial value is of type `type(0)`, i.e. `int`.

WhiteMagic:
Looking back at the code I think handing this more explicitly is a better idea. The code really only should override the given values if, for some reason, an actual incompatible data type is being provided, say `list` instead of `int`. A solution I can think of is to specify the actual type the data should be in and then attempt to cast it, and only if that cast fails to use the default values (with a cast as well). That should work around this awkward issue of things not working just because someone forgot to add a decimal place to the number.

PR #360, opened 2021-03-04
Revise default limits of Numeric variables.

### LINKED ISSUE #359: IntegerVariable and FloatVariable have surprising defaults
IntegerVariable and FloatVariable have surprising defaults limits of [0, 10] and [-1, 1], respectively. I suggest not restricting these by default. If a plugin author has particular limits in mind, they can set them -- otherwise why not let the user choose freely?

henniss:
Fixes https://github.com/WhiteMagic/JoystickGremlin/issues/359 and https://github.com/WhiteMagic/JoystickGremlin/issues/358

### DIFF
--- a/gremlin/user_plugin.py
+++ b/gremlin/user_plugin.py
@@ -24,6 +24,7 @@ import os
 import random
 import string
 import uuid
+import sys
 
 from PyQt5 import QtCore, QtGui, QtWidgets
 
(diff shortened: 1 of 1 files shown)


henniss:
Closing this PR, as upon reflection it doesn't fully solve #358 (With this, and int as initial value gets ignored). I'll fix that and send a new PR in a day or two.

PR #375, opened 2021-08-30
Added scroll bar to the user plugin variable list.

Moondragon85:
I am currently working on a user plugin that has gotten a bit out of hand. It has so many input and other variables that the list doesn't fit on my screen any longer. The only way to access the lower parts (apart from editing the xml file manually) was to maximize the window which squishes the list vertically but then the individual variables were barely readable. That's why I added a scroll bar to that list.

### DIFF
--- a/gremlin/ui/user_plugin_management.py
+++ b/gremlin/ui/user_plugin_management.py
@@ -107,6 +107,7 @@ class ModuleManagementController(QtCore.QObject):
 
         layout = self.view.right_panel.layout()
         gremlin.ui.common.clear_layout(layout)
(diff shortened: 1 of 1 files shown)


WhiteMagic:
This is something that definitely should be there, though I'm unlikely to merge this at this stage, mainly for the reason that every single UI related code piece is being redone and existing code can't be ported since I'm moving from raw Qt to QML for the frontend itself.

Looking over the code nothing really stands out other than I usually use the enums Qt defines for various settings, such as Qt::ScrollBarAlwaysOff instead of 1 in the scrollbar. That way it's easier for someone else to figure out what that magic value does.

Issue #365, opened 2021-05-01
FloatVariable does not allow negative values


edwardwbarber:
The Joystick Gremlin GUI does not seem to allow negative values to be entered for FloatVariable. Nor are they allowed for as default values:

```python
testfloat = FloatVariable(
    "Test float",
    "Test float",
    -0.1,
    -1.0,
    1.0
)
```

Output of above defaults to `0.0`. Expected `-0.1`. Attempts to enter `-0.1` manually into GUI widget are reset to `0.0`.

Looks like `QtWidgets.QDoubleSpinBox.Maximum` and `QtWidgets.QDoubleSpinBox.Minimum` aren't assigned to the limits given in `FloatVariable`. Seems like this could be fixed by inserting the following after line 345 in `gremlin/user_plugin.py` (haven't been able to create a 32-bit python 3.4 env to build JoystickGremlin or I would test it myself...):

```python
value_widget.setMaximum(float(self.max_value))
value_widget.setMinimum(float(self.min_value))
```

This might also avoid some of the confusion raised in #359. As it is the GUI lets a user enter up to 99.99 as a float, only to be ignored under the hood :)

Cheers, 
-Edward

WhiteMagic:
Fixed with 37ba776.

PR #376, opened 2021-09-01
fixed user plugin variables

Moondragon85:
Setting the initial values as well as the min and max ranges for the user plugin variables did not work. I fixed that and also increased the precision of the float variables by one decimal.

### DIFF
--- a/gremlin/user_plugin.py
+++ b/gremlin/user_plugin.py
@@ -342,6 +342,8 @@ class NumericalVariable(AbstractVariable):
             )
         elif self.variable_type == common.PluginVariableType.Float:
             value_widget = QtWidgets.QDoubleSpinBox()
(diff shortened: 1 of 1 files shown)


Moondragon85:
After reading related issue #358 I now understand what _init_numeric is supposed to do. I'll revert those changes. The rest of the fixes should be sufficient.

Moondragon85:
Done. I guess this should also fix issues #358, #359 and #365.

WhiteMagic:
Thanks for the change, in my view it doesn't fully fix #358 and #359 though it makes it possible to actually deal with them. Issue #358 is a rather simple fix, i.e. not doing a hard type check but conversion check and #359 is more of a preference/philosophy issue.

Issue #361, opened 2021-03-05
Bug: Activation Condition Comparison in XML overwritten when loaded in UI


VegaAntares:
Hello,

I found an annoying bug but hopefully easy to fix. The problem is the "comparison" item is set to "" (empty string) in the concrete classes (base_classes.VJoyCondition in my example) since the self.comparison is inherited from AbstractCondition that initialize it this default value because I guess correct initialization value depends of the InputType

So in my example such xml shows "pressed" in the comparison drop down:
```
   <button description="" id="5"">
      <container type="basic"><br/>
          <action-set>
              <map-to-keyboard>
                  <key extended="False" scan-code="37"/>  
                  <activation-condition rule="all">  
                      <condition comparison="released" condition-type="vjoy" id="31" input="button" vjoy-id="1"/>  
                  </activation-condition>  
              </map-to-keyboard> 
          </action-set>  
      </container> 
  </button>
```

Then VJoyConditionWidget.__init__ receive the correct condition_data argument 

```
class VJoyConditionWidget(AbstractConditionWidget):
   def __init__(self, condition_data, parent=None):
        # ...
        super().__init__(condition_data, parent)   # <- Set the comparison value
        # ...
        self._modify_vjoy(self.vjoy_selector.get_selection()) #<- override the value
```

So you should then only override the value if self.condition_data.comparison is an empty string
```
    if data["input_type"] == InputType.JoystickAxis:
        self.condition_data.comparison = "inside"
    elif data["input_type"] == InputType.JoystickButton:
        self.condition_data.comparison = "pressed"
    elif data["input_type"] == InputType.JoystickHat:
        self.condition_data.comparison = \
              util.hat_tuple_to_direction((0, 0))
```

And the same for all other condition Widgets
Of course this bug is not seen when one use already the default value in the XML

WhiteMagic:
This bug has been in there for a bit and I think only affects vJoy conditions, which is why it's not commonly encountered. There might already be an issue for this though I couldn't find it when looking just then. The fix for this bug is going to be that this entire piece of code won't exist anymore in the next release as the entire condition stuff, just like just about everything else, is being redone.

VegaAntares:
Yes only vJoy.
Actually this feature is pretty neat since I use the vJoy buttons as boolean for keeping track of states. I mean I do not remap them into the game but actions triggeted may depends on their state. This avoids some python plugins I made initially.
[...]

PR #407, opened 2022-07-10
Fix vjoy activation condition handling

edwardwbarber:
This has come up in the discord a couple of times (and mentioned here as well #361). Decided to take a look myself to fix the issue. This fixed the problem on my end. Happy to provide a compiled win x64 binary if that is of use too.

Here are the changes from the commit itself:

Previously all vjoy activation conditions would be reset to the default value when the UI was refreshed. This was due to several lines in `VJoyConditionWidget._modify_vjoy` which were intended to reset to a default value if the VJoy input was modified. Unfortunately, `_modify_vjoy` is called every time the UI is reset, thus resetting the condition trigger in addition to the intended
effect of updating the selected VJoy input.

To avoid the reset, we simply removed the lines that change the activation condition within `_modify_vjoy`.

### DIFF
--- a/gremlin/ui/activation_condition.py
+++ b/gremlin/ui/activation_condition.py
@@ -624,12 +624,38 @@ class VJoyConditionWidget(AbstractConditionWidget):
         self.condition_data.vjoy_id = data["device_id"]
         self.condition_data.input_type = data["input_type"]
         self.condition_data.input_id = data["input_id"]
+        
+        valid_axis_condition = [
+            "inside",
+            "outside"
+            ]
+        valid_button_condition = [
+            "pressed",
+            "released"
+            ]
+        valid_hat_condition = [
+            "center",
+            "north",
+            "north-east",
+            "east",
+            "south-east",
+            "south",
+            "south-west",
+            "west",
+            "north-west"
+            ]
 
-        if data["input_type"] == InputType.JoystickAxis:
+        if (data["input_type"] == InputType.JoystickAxis and
+            self.condition_data.comparison not in valid_axis_condition
+           ):
             self.condition_data.comparison = "inside"
-        elif data["input_type"] == InputType.JoystickButton:
+        elif (data["input_type"] == InputType.JoystickButton and
+              self.condition_data.comparison not in valid_button_condition
+             ):
             self.condition_data.comparison = "pressed"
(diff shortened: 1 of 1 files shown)


WhiteMagic:
While the fix will solve one issue I'm unclear whether it will cause another one. The true underlying issue is that the modify function is used both for initialization, where the defaulting of variable states is important, and updating whenever a vJoy value changes, where this defaulting is not intended.

Looking a bit more at the code, outright removal of the input type based checks will cause bugs. The reason for that code is to enforce the `comparison` value to have a valid value as it is possible to change the vJoy input type from button to axis or similar. Not changing the input type of the `comparison` value will break other UI aspects and likely cause crashes. Thus, the actual fix would need to check whether the `comparison` value needs to be defaulted due to an input type change on the vJoy side.

Also, the `_modify_vjoy` method isn't technically called upon a UI reset, it causes one, but rather when the vJoy selection changes, which can also happen during a UI reset as there is some amount of cyclical stuff in the UI aspect going on.

Lastly, it is not likely that any of these fixes will be merged as R14 does not reduce any of the UI, and most if not all of the action logic has to be redone as well. While fixing some of the remaining common bugs would be nice, integrating, testing, etc. of those would eat whatever little time I have to work on R14 which is why I'm unlikely to  do any of that.

edwardwbarber:
You are right. I did not consider the `comparison` value state if the input type is changed but not the comparison selection within the UI. Reverted my change and added some logic such that the `comparison` value is updated only if the current `comparison` value does not match an expected one for the type. A little clunky, but it seems to work.

I understand that you don't have the time to integrate these changes yourself, but I'd still hope this fix is useful to folks in the community that are willing to build their own versions (few of us though there may be!). Thanks for your efforts all the same.

WhiteMagic:
This looks good, there are probably some ways to make it less clunky but really it would need proper enums for the various conditions and inputs, things that don't exist in that code base, at least not for the conditions. So this is a good middle ground.

Yeah, that is why I comment and leave the pull requests open so that people actually can see them for their own use. Also helps me to remember things that I need to keep in mind when implementing those functions. This kind of bug and annoyance with condition types for example doesn't exist anymore in R14 conditions.

WhiteMagic:
Closing as R14 has been released.

PR #492, opened 2023-06-05
Fix vJoy activation issues

leon-gh:
* Fixes the issue with Activate if vJoy button is 'Released'
* Fixes the issue with Activate if vJoy axis is 'Outside'
* Fixes the issue with Activate if vJoy hat anything but 0,0


WhiteMagic:
This won't fix the problem as the conditions can now be in invalid states. A more complete fix is this one here https://github.com/WhiteMagic/JoystickGremlin/pull/407 but as stated there I won't be integrating these as all that code disappears with R14.

leon-gh:
@WhiteMagic - No issues, it works for me and has not caused any invalid states. I would still suggest to integrate either in master branch (which is r13 AFAK) and re-release JG so all people benefit from it.

Issue #425, opened 2022-08-20
Action library interface


WhiteMagic:
R14's profile structure no longer assigns actions directly to inputs but rather assigns library items to actions. The library items are the objects that encode the actions to run. Therefore the same actions can easily be shared between different inputs or moved from one to the other.

Currently, the profile structure is implemented and the UI creates the library items and assigns them to actions. However, the UI to manage and use the library directly is missing. The UI will need to provide the following functionality:

- Add library items directly to suitable inputs
- Change the description associated with the item
- List where and by how many inputs the item is being used
- Allow for removal and correct update of the entire profile

Related issues: #179, #172, #102, #170

WhiteMagic:
- Adding items to actions/inputs will be done via the Reference action
- Removing items will not be supported as that can lead to highly undesirable outcomes due to connections between actions not being obvious
- Showing usage numbers and modifying descriptions can be a separate management UI that will be added at a later point

PR #514, opened 2023-12-29
Added Clipboard implementation

leon-gh:
* Supports copy/pasting actions from one button to another.

Please consider such functionality (if not already) for r14. Feedback is welcome (NOTE: I don't consider myself a Qt developer at all, so I might have done things at wrong places).

The feature is very much useful when moving things around. I have more ideas on right-clicking the button and giving a user ability to:
- move actions to another mode
- copy actions from another mode
- copy actions to another mode

But all of these could use Clipboard to get stuff done.

### DIFF
--- a/gremlin/clipboard.py
+++ b/gremlin/clipboard.py
@@ -15,30 +15,23 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import gremlin.actions
-import gremlin.base_classes
-import gremlin.cheatsheet
-import gremlin.code_runner
-import gremlin.common
-import gremlin.config
-import gremlin.control_action
-import gremlin.error
-import gremlin.event_handler
-import gremlin.execution_graph
-import gremlin.fsm
-import gremlin.hid_guardian
-import gremlin.hints
-import gremlin.input_devices
-import gremlin.joystick_handling
-import gremlin.macro
-import gremlin.plugin_manager
-import gremlin.process_monitor
-import gremlin.profile
-import gremlin.repeater
-import gremlin.shared_state
-import gremlin.sendinput
-import gremlin.spline
-import gremlin.tts
-import gremlin.util
-import gremlin.windows_event_hook
-import gremlin.clipboard
+from . import common
+
+
+@common.SingletonDecorator
+class Clipboard:
+
+    """Responsible for storing and restoring action related data."""
+
+    def __init__(self):
+        """Creates a new instance, initializing empty clipboard."""
+        self.item_data = None
+
+    def copy(self, item_data):
+        """ Saves item_data to clipboard """
+        self.item_data = item_data
+
+    def paste(self):
+        """ Returns item_data saved in clipboard """
+        return self.item_data
+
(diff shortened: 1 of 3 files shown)


leon-gh:
@WhiteMagic - I know r13 is frozen, I opened PR just to get feedback and if you see things changing for r14 a lot that such approach could not be used.

WhiteMagic:
R14 has an entirely different profile structure which differs drastically from R13 and prior. An issue with the clipboard as implemented here is that it is not granular enough. It will copy an entire input rather than particular actions or even containers, so I wouldn't go for a solution like this.

Configurations in R14 are split differently. In R13 actions are directly linked to inputs, this changes in R14 where actions are stored in a shared library and inputs simply refer to a set of actions they use. This makes it possible to add existing actions to new inputs, which is effectively copy&paste. Something to work out is whether those assignments are treated as copies or references. It would also be possible, though not something I'm considering currently, to add copy and paste as actions or similar to individual actions. All this is significantly simpler in R14 as the information required for these operations is just an identifier rather than actual data, which is already used by the drag&drop part of the UI.

#425 has some pointers to other issues around this whole idea of making it easier to duplicate and move configuration aspects.

332fg-raven:
@WhiteMagic - thanks for the feedback. It's good to hear that R14 will be including these kind of features, which make profile editing more efficient.
This copy/paste was designed to copy a full container, rather than individual pieces - use case for this was I configured "a sophisticated" container, then realized it's in the wrong mode. I just wanted to copy it to another mode. Ended up editing xml file to move things around. I understand and see the use of actions copying too, I think R14 can understand where the CTRL+C was pressed - if object focused was a button (left pane), then it could copy a whole container, if it was an action inside or selected actions, then it would copy just that portion.

Thanks!

leon-gh:
@WhiteMagic - thanks for the feedback. It's good to hear that R14 will be including these kind of features, which make profile editing more efficient.
This copy/paste was designed to copy a full container, rather than individual pieces - use case for this was I configured "a sophisticated" container, then realized it's in the wrong mode. I just wanted to copy it to another mode. Ended up editing xml file to move things around. I understand and see the use of actions copying too, I think R14 can understand where the CTRL+C was pressed - if object focused was a button (left pane), then it could copy a whole container, if it was an action inside or selected actions, then it would copy just that portion.

Thanks!

WhiteMagic:
In the new system there no longer is the distinction between container and actions. A container is effectively just an action that has other actions as children. So copying an action will actually copy everything "associated" with it. So it could be a simple remap or a highly complicated tempo container with conditions, macros, etc.

Figuring out the UI side of things will be more challenging, as just keyboard shortcuts won't know what part the user is interested in. One option is to have a context menu or copy&paste buttons on the UI, thought that might get a bit unwieldy.

WhiteMagic:
Linking and duplicating of actions are supported in R14 via the `Reference` action (https://github.com/WhiteMagic/JoystickGremlin/blob/develop/action_plugins/reference/__init__.py). It does the correct thing as far as functionality goes but can be a bit clunky to figure out which is the "right" action. Though an additional interface that allows picking of actions more cleanly is something to be added in the future.

PR #527, opened 2024-03-29
Implement bi-directional axis merge operation

Drakehinst:
Fix of the merge axis average formula (it was equivalent an unbounded sum).

The formula is now explained for added clarity.

### DIFF
--- a/action_plugins/merge_axis/__init__.py
+++ b/action_plugins/merge_axis/__init__.py
@@ -19,7 +19,7 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import List, NamedTuple, Optional
+from typing import List, NamedTuple, Optional, Any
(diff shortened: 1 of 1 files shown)


WhiteMagic (changes requested):
I'd simplify the description of the _bidirectional operation to explain only the intent and not the full math as that should be straight forward enough to understand when the desired behavior is understood. Maybe something like this.


```Merges two axes into one, using the first axes to control the lower half of the value range and the second axis for the upper half. Allows full deflection either way if only one axis is used and differential mixing if both are used. An example use case of this is merging two axis on pedals into a single left/right axis.```

Drakehinst:
Sounds good 👍

WhiteMagic (approved):
Looks good!

PR #560, opened 2024-10-15
Support for profile loading

leon-gh:
* This feature is a hack, but allows buttons to be configured to load another profile.

I know R13 is frozen, but just submitting it to keep the idea on on the radar for R14.


WhiteMagic:
If you're interested, you should actually be able to implement this rather quickly against R14 on the develop branch via the ui.backend class. It might introduce a cyclic import issue possibly which would have to be worked around.

The develop branch now also supports defining a folder from which Gremlin will also load other plugins so not every plugin has to be a part of the main release, so that could also be used in that way.

leon-gh:
I'll try. I've got develop branch running, but need to learn the QT to understand how to work with qml - no Play Sound action yet implemented in develop branch.

leon-gh:
@WhiteMagic can you please check https://github.com/leon-gh/JoystickGremlin/tree/support-for-load-profile and tell me if I'm headed in the right direction. The UI part seems to be working OK, now just need to study the ui.backend and integrate it so the profile gets actually activated.

WhiteMagic:
Do you want to turn that into a pull request, that way it would be much easier to add comments directly at the right places in the code rather than going off lines etc.

leon-gh:
PR opened for develop branch. Closing this PR.

PR #561, opened 2024-10-18
Initial skeleton for profile loading support

leon-gh:
Currently just a placeholder for selecting a filename. Debug log is displayed when action is invoked. Next step is to integrate it with ui.backend to perform the actual profile loading.

### DIFF
--- a/action_plugins/load_profile/LoadProfileAction.qml
+++ b/action_plugins/load_profile/LoadProfileAction.qml
@@ -1,0 +1,77 @@
+// -*- coding: utf-8; -*-
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import QtQuick.Window
+import QtQuick.Dialogs
+
+import QtQuick.Controls.Universal
+
+import Gremlin.Profile
+import Gremlin.ActionPlugins
+import "../../qml"
+
+
+Item {
+    property LoadProfileModel action
+
+    implicitHeight: _content.height
+
+    RowLayout {
+        id: _content
+
+        anchors.left: parent.left
+        anchors.right: parent.right
+
+        Label {
+            id: _label
+
+            Layout.preferredWidth: 150
+
+            text: "Profile filename"
+        }
+
+        TextField {
+            id: _profile_filename
+
+            Layout.fillWidth: true
+
+            placeholderText: null != action ? null : "Enter a profile filename"
+            text: action.profile_filename
+            selectByMouse: true
+
+            onTextChanged: {
+                action.profile_filename = text
+            }
+        }
+
+        Button {
+            text: "Select File"
+            onClicked: fileDialog.open()
+        }
+
+   }
+
+   FileDialog {
+        id: fileDialog
+        title: "Select a File"
+        onAccepted: {
+            _profile_filename.text = new URL(selectedFile).pathname.substring(1)
+        }
+    }
+}
--- a/action_plugins/load_profile/LoadProfileAction.qml
+++ b/action_plugins/load_profile/LoadProfileAction.qml
@@ -47,7 +47,7 @@ Item {
         }
 
         TextField {
-            id: profile_filename
+            id: _profileFilename
 
             Layout.fillWidth: true
 
@@ -62,16 +62,16 @@ Item {
 
         Button {
             text: "Select File"
-            onClicked: fileDialog.open()
+            onClicked: _fileDialog.open()
         }
    }
 
    FileDialog {
-        id: fileDialog
+        id: _fileDialog
         nameFilters: ["Profile files (*.xml)"]
         title: "Select a File"
         onAccepted: {
-            profile_filename.text = selectedFile.toString().substring("file:///".length)
+            _profileFilename.text = selectedFile.toString().substring("file:///".length)
         }
     }
 }
--- a/action_plugins/load_profile/LoadProfileAction.qml
+++ b/action_plugins/load_profile/LoadProfileAction.qml
@@ -47,7 +47,7 @@ Item {
         }
 
         TextField {
-            id: _profile_filename
+            id: profile_filename
 
             Layout.fillWidth: true
 
@@ -64,14 +64,14 @@ Item {
             text: "Select File"
             onClicked: fileDialog.open()
         }
-
    }
 
    FileDialog {
         id: fileDialog
+        nameFilters: ["Profile files (*.xml)"]
         title: "Select a File"
         onAccepted: {
-            _profile_filename.text = new URL(selectedFile).pathname.substring(1)
+            profile_filename.text = selectedFile.toString().substring("file:///".length)
         }
     }
 }
--- a/action_plugins/load_profile/__init__.py
+++ b/action_plugins/load_profile/__init__.py
@@ -1,0 +1,173 @@
+# -*- coding: utf-8; -*-
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from __future__ import annotations
+
+import os
+import logging
+from typing import List, TYPE_CHECKING
+from xml.etree import ElementTree
+
+from PySide6 import QtCore
+from PySide6.QtCore import Property, Signal
+
+from gremlin import event_handler, util
+from gremlin.base_classes import AbstractActionData, AbstractFunctor, \
+    Value
+
+from gremlin.profile import Library
+from gremlin.types import ActionProperty, InputType, PropertyType
+
+from gremlin.ui.action_model import SequenceIndex, ActionModel
+from gremlin.ui import backend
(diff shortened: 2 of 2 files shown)


action_plugins/load_profile/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    ]
+
+    def __init__(
+            self,
+            behavior_type: InputType=InputType.JoystickButton
+    ):
+        super().__init__(behavior_type)
+
+        # Model variables
+        self.profile_filename = ""
```
WhiteMagic, line 135: Might be worth to use a `pathlib.Path` object here so you can easily verify the provided string is valid and possibly also use that for the valid aspect. The `is_valid()` bit is mainly there to avoid saving "bad" configuration bits.

action_plugins/load_profile/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    tag = "load-profile"
+    icon = "\uF3B9"
+
+    functor = LoadProfileFunctor
+    model = LoadProfileModel
+
+    properties = [
+        ActionProperty.ActivateOnBoth
+    ]
+    input_types = [
```
WhiteMagic, line 123: This should only be for `JoystickButton` and `Keyboard` as they are the only two "push" style inputs. The other two inputs will show this action dynamically when they are switched to behave like that via the UI.

action_plugins/load_profile/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+    version = 1
+    name = "Load Profile"
+    tag = "load-profile"
+    icon = "\uF3B9"
+
+    functor = LoadProfileFunctor
+    model = LoadProfileModel
+
+    properties = [
```
WhiteMagic, line 119: You probably don't want this, as that would trigger a profile load when the input is pressed and again when released. The better one is `ActivateOnPress` as that only is the default behavior not the one that it is restricted to. You also may want to add the `AlwaysExecute` property as otherwise you won't be able to load a profile when the current profile is paused. In R14 that "run when paused" is a per-action-class setting rather than per-instance as was the case in R13.

action_plugins/load_profile/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+        self._profile_filename = action.profile_filename
+
+    def __call__(
+        self,
+        event: event_handler.Event,
+        value: Value
+    ) -> None:
+
+        if not self._should_execute(value):
+                return
```
WhiteMagic: That indentation seems too deep

action_plugins/load_profile/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class LoadProfileFunctor(AbstractFunctor):
+
+    """Executes a load profile action callback."""
+
+    def __init__(self, action: LoadProfileData):
+        super().__init__(action)
+
+        self._profile_filename = action.profile_filename
```
WhiteMagic: No real need to store this, as everything is stored in the data object which is stored as `self.data`. This has the nice side-effect that changing of values while the profile is active are automatically reflected in the running profile.

action_plugins/load_profile/LoadProfileAction.qml
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        Button {
+            text: "Select File"
+            onClicked: fileDialog.open()
+        }
+
+   }
+
+   FileDialog {
```
WhiteMagic, line 69: Technically this should need `import Qt.labs.platform` but maybe through the way that the UI is constructed this is implicitly available.

WhiteMagic: Would be a good idea to make use of the filter function of this to only show files that should be loadable, i.e. XML files.

https://doc.qt.io/qt-6/qml-qt-labs-platform-filedialog.html#nameFilters-prop

action_plugins/load_profile/LoadProfileAction.qml
```diff
@@ -0,0 +1,77 @@
...
+        id: fileDialog
+        title: "Select a File"
+        onAccepted: {
+            _profile_filename.text = new URL(selectedFile).pathname.substring(1)
```
WhiteMagic: - Might also just use `file` here as that's the value of the path selected upon acceptance while `selectedFile` is constantly changing.
- Does this return the complete absolute path to the selected string?

WhiteMagic:
Left you a few comments but otherwise looks like there is nothing that seriously is broken with it. There is also a `qml_notes.md` document in the root directory which contains a bunch of things I've learned about QML and its quirks. This book is also somewhat useful to get an idea of some of QML's aspects https://www.qt.io/product/qt6/qml-book

action_plugins/load_profile/__init__.py
leon-gh: I implemented is_valid() using os.isfile and os.access. If you prefer pathlib.Path() then the profile_filename will become None at init and object when assigned. Let me know.

action_plugins/load_profile/__init__.py
leon-gh: Noted.

action_plugins/load_profile/__init__.py
leon-gh: Will fix the indent and other unused imports as a part of cleanup.

action_plugins/load_profile/__init__.py
leon-gh: OK.

action_plugins/load_profile/LoadProfileAction.qml
leon-gh: FileDialog returns selectedFile as url (file:///C:/path/file). 
I converted it to string, then used .substring("file:///".length)) on it. If not OK, please advise how do you want me to change it.

leon-gh:
> Left you a few comments but otherwise looks like there is nothing that seriously is broken with it. There is also a `qml_notes.md` document in the root directory which contains a bunch of things I've learned about QML and its quirks. This book is also somewhat useful to get an idea of some of QML's aspects https://www.qt.io/product/qt6/qml-book

Yeah, went through this book by looking at what I needed :-). What I miss is ability to debug qml somehow. If you can debug QML it would be nice to have the same IDE configurations.

action_plugins/load_profile/LoadProfileAction.qml
leon-gh: FileDialog needs import QtQuick.Dialogs. Added filters for *.xml (had plans, but just didn't make it in yet :-))

action_plugins/load_profile/__init__.py
leon-gh: OK. What about JoystickHat? Is it not a push style input?

action_plugins/load_profile/__init__.py
WhiteMagic: I think both are fine, in most of the new parts of Gremlin I've starting using `Path` objects instead of strings and `os.` calls.

action_plugins/load_profile/__init__.py
WhiteMagic: A hat reports 9 values, directions & center, but only gives you a "new value" not a "let go of this direction". One can map that onto button-behavior which Gremlin does when the hat is treated as a button in Gremlin. So actions that require a "press" action only make sense for the two button-like inputs though others can be made to behave as such.

WhiteMagic:
Yeah ... debugging QML is not something I ever figured out if you can really do. It's really all kinds of trying things out and making stuff visual, so the worst kind of printf debugging.

action_plugins/load_profile/LoadProfileAction.qml
```diff
@@ -0,0 +1,77 @@
...
+        Label {
+            id: _label
+
+            Layout.preferredWidth: 150
+
+            text: "Profile filename"
+        }
+
+        TextField {
+            id: profile_filename
```
WhiteMagic: Just as an FYI I tend to do `_camelCase` for IDs (semi-successful at remembering this myself) as Qt and QML use that which is also why you'll see slots use camel case for method names. And I add the underscore so I can easily recognize an ID reference in some of the bigger QML files.

action_plugins/load_profile/__init__.py
```diff
@@ -0,0 +1,171 @@
...
+    ) -> None:
+
+        if not self._should_execute(value):
+            return
+
+        logging.getLogger("system").debug(
+            f"Loading profile ...{self.data.profile_filename}"
+        )
+
+        be = backend.Backend()
```
WhiteMagic, line 62: Maybe worth checking the file is still valid here or in the `_from_xml()` (likely better since I hope people don't move a profile when they actively use Gremlin) call to prevent attempting to load an inaccessible one. I think you should be fine with regards to profile errors as that error chain leads to a UI error dialogue. Starting can throw exceptions and they might be uncaught until termination possibly.

action_plugins/load_profile/LoadProfileAction.qml
WhiteMagic: As long as you get the valid absolute path out of it I would say that's what you want. Was mostly wondering, as I don't remember anymore what those calls spat out.

action_plugins/load_profile/__init__.py
leon-gh: OK, then I'll make it consistent, no issues.

action_plugins/load_profile/__init__.py
leon-gh: OK, understood.

action_plugins/load_profile/LoadProfileAction.qml
leon-gh: Got it, n/p.

action_plugins/load_profile/LoadProfileAction.qml
leon-gh: OK, will update.

action_plugins/load_profile/__init__.py
```diff
@@ -133,24 +132,27 @@ def __init__(
         super().__init__(behavior_type)
 
         # Model variables
-        self.profile_filename = ""
+        self._profileFilename = ""
```
WhiteMagic: Ah sorry I think I confused you, the only place I use non-pythonic camelCase variables is QML exposed things, in the data class and really everywhere else it's normal Python convention, i.e. `self.profile_name` for this one here unless you intend to expose it via a property.

action_plugins/load_profile/__init__.py
leon-gh: Added validator to _from_xml()

action_plugins/load_profile/__init__.py
leon-gh: Turns out pathlib.Path code lacks .access() to validate whether file is accessible. Left os methods since Path code turned out more lines and more complex.

action_plugins/load_profile/__init__.py
leon-gh: n/p, I thought *Data class was directly linked to QML, but it's via *Model getter/setters - new to Qt/QML.

WhiteMagic:
Awesome work, thanks.

PR #565, opened 2024-10-22
Play Sound action support

leon-gh:


action_plugins/play_sound/__init__.py
```diff
@@ -0,0 +1,183 @@
...
+    def _get_sound_volume(self) -> int:
+        return self._data.sound_volume
+
+    def _set_sound_volume(self, value: int) -> None:
+        if value == self._data.sound_volume:
+            return
+        self._data.sound_volume = value
+        self.soundVolumeChanged.emit()
+
+    sound_filename = Property(
```
WhiteMagic: I would name them `soundFilename` and `soundVolume` as they are QML properties and that's their preferred variable naming convention. Essentially things I expose/use in QML follows their convention everything else is normal python.

action_plugins/play_sound/__init__.py
```diff
@@ -0,0 +1,183 @@
...
+    name = "Play Sound"
+    tag = "play-sound"
+    icon = "\U0001F39C"
+
+    functor = PlaySoundFunctor
+    model = PlaySoundModel
+
+    properties = [
+        ActionProperty.ActivateOnPress,
+        ActionProperty.AlwaysExecute
```
WhiteMagic: Not sure it's necessary to be able to play sounds when a profile is paused, i.e. not meant to react.

action_plugins/play_sound/__init__.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def __init__(
+            self,
+            behavior_type: InputType = InputType.JoystickButton
+    ):
+        super().__init__(behavior_type)
+
+        # Model variables
+        self.sound_filename: str = ""
+        self.sound_volume: int = 50
```
WhiteMagic, line 145: You don't seem to be (de)serializing the sound volume.

gremlin/audio_player.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from PySide6 import QtCore
+from PySide6.QtCore import QUrl
+
+from gremlin.common import SingletonDecorator
+from PySide6.QtMultimedia import QAudioOutput, QMediaPlayer
```
WhiteMagic: I usually order imports from system -> external libraries -> gremlin things.

gremlin/audio_player.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self) -> None:
+        QtCore.QObject.__init__(self)
+        self.player_output = QAudioOutput()
+        self.player = QMediaPlayer()
+        self.player.setAudioOutput(self.player_output)
+
+    def play(self, sound_filename: str, volume: int) -> None:
+        """ volume: 0 = mute, >0 = volume level, max volume is 100 """
+
+        if self.player.playbackState() != QMediaPlayer.StoppedState:
```
WhiteMagic: Does this mean triggering multiple sounds in quick succession, i.e. before one is finished, the already playing one will be interrupted and the new one play? This may not be desirable as people could expect them to queue.

Maybe a nice workaround is to implement queuing and priority levels.
- same priority level: queue the new sound and wait for the old one to finish
- higher priority level: stop currently playing sound and play the higher importance one

In that manner a user could actually have different priorities to control the "interruption" behavior and otherwise have them queue. We could then also add an option for the play sound action that lets the user define if they want queuing or overriding for same priority sounds.

action_plugins/play_sound/PlaySoundAction.qml
```diff
@@ -0,0 +1,94 @@
...
+            text: "Select File"
+            onClicked: _fileDialog.open()
+        }
+
+        Label {
+            Layout.preferredWidth: 50
+            text: "Volume"
+        }
+
+        TextField {
```
WhiteMagic: I would use a SpinBox (https://doc.qt.io/qt-6/qml-qtquick-controls-spinbox.html) here as the input is meant to be an integer. That widget lets you set minimum and maximum values. The `MapToMouseAction.qml` file is a decent example of how to use that widget.

leon-gh: Yup, no issues.

action_plugins/play_sound/__init__.py
leon-gh: Got it.

action_plugins/play_sound/__init__.py
leon-gh: Copy-paste leftover. Will update.

gremlin/audio_player.py
leon-gh: n/p :-)

action_plugins/play_sound/__init__.py
leon-gh: Good catch! :-)

gremlin/audio_player.py
leon-gh: As agreed added a global Action Play Sound configuration to allow sequential playing. Feel free to suggest better descriptions of buttons/texts. Priorities were not added and can be later, if requested.

action_plugins/play_sound/PlaySoundAction.qml
```diff
@@ -69,16 +69,15 @@ Item {
...
-
-            Layout.fillWidth: false
-            Layout.preferredWidth: 50
-            text: action.sound_volume
-            selectByMouse: true
-
-            onTextChanged: {
-                action.sound_volume = text
+            Layout.preferredWidth: 100
+            value: 50
```
WhiteMagic: This won't show the correct value that the action has stored, you need to use the property binding stuff from QML to have the UI reflect the model's state. Though in your case that's really easy

`value: action.sound_volume`

action_plugins/play_sound/__init__.py
```diff
@@ -141,28 +141,36 @@ def __init__(
         super().__init__(behavior_type)
 
         # Model variables
-        self.sound_filename: str = ""
-        self.sound_volume: int = 50
+        self.soundFilename: str = ""
```
WhiteMagic: For consistency with the rest of Gremlin it would be preferrable for these to be `sound_filename` and `sound_volume`

action_plugins/play_sound/__init__.py
```diff
@@ -141,28 +141,36 @@ def __init__(
...
+                ["filename", self.soundFilename, PropertyType.String],
+                ["volume", self.soundVolume, PropertyType.Int],
+            ]
+        )
         return node
 
     def is_valid(self) -> bool:
-        if len(self.sound_filename) > 0 and os.path.isfile(self.sound_filename) and \
-           os.access(self.sound_filename, os.R_OK):
+        if len(self.soundFilename) > 0 and os.path.isfile(self.soundFilename) and \
```
WhiteMagic: If you think this "does that file exist and can I access it" might be useful more generally, feel free to extract it out into a function in `gremlin.utils` as you've also needed this functionality in the load plugin action.

action_plugins/play_sound/__init__.py
WhiteMagic: Oh I think I confused you as you seem to have flipped them around of what I meant to describe :-)

gremlin/audio_player.py
```diff
@@ -14,31 +14,66 @@
...
     def play(self, sound_filename: str, volume: int) -> None:
         """ volume: 0 = mute, >0 = volume level, max volume is 100 """
 
-        if self.player.playbackState() != QMediaPlayer.StoppedState:
+        sequential_play = Configuration().value("action", "play-sound", "sequential-play")
+        if (
+            sequential_play and self.player.playbackState() != QMediaPlayer.StoppedState
+        ):
+            playlist_item: PlayListItem = PlayListItem(sound_filename, volume)
+            self.play_list.append(playlist_item)
```
WhiteMagic: Probably find to just in place construct and append the item.

gremlin/ui/backend.py
```diff
@@ -113,6 +114,8 @@ def activate_gremlin(self, activate: bool):
         else:
             # Stop running the code
             self.runner.stop()
+            player = AudioPlayer()
```
WhiteMagic: I usually do `Singleton().method()`, i.e. `AudioPlayer().stop()` for these things when I know that I don't need the thing again.

leon-gh: n/p

gremlin/audio_player.py
leon-gh: n/p

action_plugins/play_sound/__init__.py
leon-gh: Good idea. Since I've used the same piece in Load Profile I'd prefer doing it as a separate PR. OK?

action_plugins/play_sound/__init__.py
leon-gh: n/p, I think I got better understanding now how QML action properties and data class work.

action_plugins/play_sound/__init__.py
leon-gh: yup :-), n/p, got it now.

action_plugins/play_sound/PlaySoundAction.qml
leon-gh: Yup, thanks - this was a left over from the first implementation.

WhiteMagic (approved):
Awesome, looks good to me with all the changes.

PR #568, opened 2024-10-29
Support for button and axis names

leon-gh:
* Created an external device_db.json file which contains mapping of buttons to their names, e.g. F-16 stick might have a Button 10 which is known as TMS UP, so Gremlin would be able to display button info as Button 10 - TMS UP
* Added configurable option to display button names as:
  - None - Button ID
  - Mixed - Button ID + Button name
  - Name - Button name


gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+@SingletonDecorator
+class DeviceDatabase:
+    """Provides button/axis to name mapping for known devices"""
+
+    def __init__(self) -> None:
+        self._load()
+
+    def _load(self) -> None:
+
+        db_file = "device_db.json"
```
WhiteMagic: Probably will need to use `util.resource_path()` here such that the file can be found when the whole thing is "compiled" with PyInstaller.

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+        json_data = {}
+        with open(db_file) as hdl:
+            try:
+                decoder = json.JSONDecoder()
+                json_data = decoder.decode(hdl.read())
+                load_successful = True
+            except ValueError:
+                pass
+
+        if not load_successful:
```
WhiteMagic: We might want to do a check on the contents to catch obvious mistakes like misspelt keys. So `button 1` doesn't work but `Button 1` does which we can easily catch, and the same if there is a mapping that is missing one end we can also emit warnings to the logfile.

This could also be done via Json Schema but I feel that this might be overkill :-)

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+        Returns the input name based on the global configuration option
+        and availability of information about the input in
+        the device database.
+
+        If a device exists in the device database and it's mapping or
+        the input are not defined, return a default
+        input name, e.g. Axis 1 or Button 1.
+
+        Some devices have configurable number of buttons and/or axes so
+        adjusting device database information might be required in the future.
```
WhiteMagic: I usually add argument and return value annotation as well in the following manner:
```
Args:
    input_name: ...
    input_map: ...

Returns:
    ...
```
This will allow sphinx to later on generate useful documentation of everything, something R13 was missing. And thankfully we don't need the weird type annotation since typing now contains that info.

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+        """
+        input_name_display_mode = Configuration().value(
+            "global", "input-names", "input-name-display-mode"
+            )
+
+        if input_name_display_mode == "None" or input_name not in input_map:
+            return input_name
+
+        db_input_name = input_map[input_name].strip()
```
WhiteMagic: This will fail if the mapping is incomplete, i.e. `input_name` does not exist in the mapping. Using `.get(input_name, "")` should work around that issue.

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+        """Returns the button/axis <-> name mapping of a device"""
+        if device is None:
+            return None
+
+        for d in self.device_db["devices"]:
+            if (
+                d["product"] == device.product_id and
+                d["vendor"] == device.vendor_id
+            ):
+                if (
```
WhiteMagic: Probably can be simplified to something like `d.get("mapping", None) in self.device_db["mapping|]` I would assume that `None` won't be found in the mapping dict.

gremlin/ui/device.py
```diff
@@ -194,6 +290,7 @@ def __init__(self, parent=None):
         super().__init__(parent)
 
         self._device: dill.DeviceSummary | None = None
+        self.device_mapping: Dict[str, str] | None = None
```
WhiteMagic: Since we don't expect to expose this I'd turn it into a private variable, i.e. `self._device_mapping`

gremlin/ui/device.py
```diff
@@ -997,4 +1100,18 @@ def axisIdentifier(self, index: int) -> int:
...
+
+Configuration().register(
+    "global",
+    "input-names",
+    "input-name-display-mode",
+    PropertyType.Selection,
+    "Mixed",
+    "Defines how input name is displayed.",
+    {
+        "valid_options": ["None", "Mixed", "Name"]
```
WhiteMagic: "None" is a somewhat confusing label as it might be interpreted as showing no labels at all. Maybe "Numerical", "Numerical & Label", "Label" might be clearer?

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+from gremlin.config import Configuration
 
 
 QML_IMPORT_NAME = "Gremlin.Device"
 QML_IMPORT_MAJOR_VERSION = 1
 
 
+@SingletonDecorator
+class DeviceDatabase:
+    """Provides button/axis to name mapping for known devices"""
```
WhiteMagic: I assume hats would be in here too as we don't restrict the code?

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+
+    def __init__(self) -> None:
+        self._load()
+
+    def _load(self) -> None:
```
WhiteMagic, line 100: Technically could be part of the constructor, though allows us to reload it in the future if needed.

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+
+        db_file = "device_db.json"
+        if not file_exists_and_is_accessible(db_file):
+            return
+
+        load_successful = False
+        json_data = {}
+        with open(db_file) as hdl:
```
WhiteMagic: Is there any benefit to this over a more concise `json_data = json.load(open(...))` in a try except block?

gremlin/ui/device.py
```diff
@@ -30,13 +31,108 @@
...
+                pass
+
+        if not load_successful:
+            logging.getLogger("system").error(
+                    f"There was an error loading {db_file}. "
+                    "Probably it's a syntax error, "
+                    "please fix the errors and restart the application.")
+
+        # create an empty device database if load was not successful
+        self.device_db = json_data if load_successful else {}
```
WhiteMagic: Does this not having "device" and "mapping" keys cause issues?

gremlin/ui/device.py
```diff
@@ -283,10 +381,15 @@ def inputIdentifier(self, index: int) -> InputIdentifier:
...
     def _name(self, identifier: Tuple[InputType, int]) -> str:
-        return "{} {:d}".format(
+        input_name = "{} {:d}".format(
             InputType.to_string(identifier[0]).capitalize(),
             identifier[1]
-        )
+            )
+
+        if self.device_mapping is not None:
+            return DeviceDatabase().input_name(input_name, self.device_mapping)
```
WhiteMagic: Since we get a mapping might be nicer for this to be an actual object we can just use rather than having to go through the database each time as it's a bit weird that we give that singleton it's own data back.

gremlin/ui/device.py
leon-gh: n/p

gremlin/ui/device.py
leon-gh: n/p

gremlin/ui/device.py
leon-gh: Line 93 makes sure that input_name is in the mapping. If you want me to refactor it into .get, no issues, but line 93 checks whether input name is in the input_map and returns a default, if not.

gremlin/ui/device.py
leon-gh: ok

gremlin/ui/device.py
leon-gh: Yup, I agree.

gremlin/ui/device.py
leon-gh: Originally I had it in the constructor, but I don't like long constructors and made it into a method. And yes, we could reload if contents changes in the future.

gremlin/ui/device.py
leon-gh: No, I usually used json.load(), but saw this method in gremlin/config.py and reused it. If json.load() is preferred, I can refactor it.

gremlin/ui/device.py
leon-gh: Good catch. Added defaults.

gremlin/ui/device.py
WhiteMagic: Ahh yeah I think I realized that the first time around and the second time I didn't see it. I think that's fine, the only thing I could think of to "disarm" that confusion is to have all that follows after in the `else` branch of the check in 93 but not sure if that makes it worse.

gremlin/ui/device.py
WhiteMagic: Yeah I feel the same, so keeping it as is sounds good to me.

gremlin/ui/device.py
WhiteMagic: That code probably was written eons ago and I wouldn't know why I decided to do it that way, so I think doing the json.load() bit is better as that's what I'd gravitate towards nowadays too.

gremlin/ui/device.py
leon-gh: Refactored into DeviceMapping object.

gremlin/ui/device.py
leon-gh: OK, will refactor.

gremlin/ui/device.py
leon-gh: ok

gremlin/ui/device.py
leon-gh: ok

gremlin/ui/device.py
leon-gh: Yes, Hats are supported too.

gremlin/ui/device.py
```diff
@@ -30,13 +32,164 @@
...
+        This piece could be refactored with JSON schema validation code,
+        if need be in the future.
+        """
+        parser_successful = True
+
+        # check top structure: devices and mappings
+        if (
+            device_db.get("mapping", None) is None or
+            device_db.get("devices", None) is None or
+            not isinstance(device_db.get("mapping", []), dict) or
```
WhiteMagic: Is this intended to be a mismatched default return type for `get` on mapping? The devices one is [] -> list while mapping is [] -> dict.

gremlin/ui/device.py
```diff
@@ -30,13 +32,164 @@
...
+        ):
+            logging.getLogger("system").error(
+                    "DeviceDatabase is corrupt and/or is missing mapping or device information.")
+            parser_successful = False
+
+        # check devices contain mandatory fields
+        device_count = 0
+        for dev in device_db.get("devices", []):
+            if (
+                dev.get("product_id", None) is None or
```
WhiteMagic: Technically this could be made a bit more concise and possibly clearer with this to not have to rely on getting `None` via default value.
`if not("product_id" in dev and "vendor_id" in dev and "mapping" in dev):`

gremlin/ui/device.py
```diff
@@ -30,13 +32,164 @@
...
+                dev.get("vendor_id", None) is None or
+                dev.get("mapping", None) is None
+            ):
+                logging.getLogger("system").error(
+                    f"DeviceDatabase device structure is corrupt for device {dev}")
+                parser_successful = False
+            else:
+                device_count += 1
+
+        mapping_count = 0
```
WhiteMagic: Would this be the same as `mapping_count = len(device_db.get("mapping", [])`?

gremlin/ui/device.py
```diff
@@ -30,13 +32,164 @@
...
+        if device is None:
+            return None
+
+        for dev in self._device_db.get("devices", []):
+            if (
+                dev.get("product_id", None) == device.product_id and
+                dev.get("vendor_id", None) == device.vendor_id
+            ):
+                if (
+                    "mapping" not in dev or
```
WhiteMagic: `"mapping" not in dev` should not occur anymore due to the `_parse_device_db` check, correct?

WhiteMagic:
Overall looks good to me, added a few small comments/clarification questions and after that this should be good in my eyes.

PR #572, opened 2024-11-08
PR 571: fix for horizontal scrolling of device list area

leon-gh:
Adding links for inspiration about potential future UI changes around Intermediate Ouput:
- height, anchors: https://stackoverflow.com/questions/28503715/why-does-adding-a-mousearea-with-anchors-fill-parent-cause-the-rows-to-stack-up
- signals & properties - https://doc.qt.io/qt-6/qtqml-syntax-signals.html & https://doc.qt.io/qt-6/qtqml-syntax-objectattributes.html

### DIFF
--- a/qml/DeviceList.qml
+++ b/qml/DeviceList.qml
@@ -46,8 +46,8 @@ Item {
         delegate: _deviceDelegate
 
         // Make it behave like a sensible scrolling container
-        ScrollBar.vertical: ScrollBar {}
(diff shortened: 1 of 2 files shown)


WhiteMagic:
Was able to work around it without signals by using TabBar (ListView effectively) and putting everything in the same UI component and fixing some other quirks, like no scrolling. See b59af92615dfe63cdc116127a03632d7d338c903

PR #607, opened 2025-04-19
Added Launch Application plugin

leon-gh:
* This plugin can launch any application as an action

### DIFF
--- a/action_plugins/launch_app/LaunchAppAction.qml
+++ b/action_plugins/launch_app/LaunchAppAction.qml
@@ -68,7 +68,7 @@ Item {
 
    FileDialog {
         id: _fileDialog
-        nameFilters: ["Application (*.bat, *.exe)"]
+        nameFilters: ["Application (*.bat, *.exe)", "All applications (*.*)"]
         title: "Select an application"
         onAccepted: {
             _applicationToLaunch.text = selectedFile.toString().substring("file:///".length)
--- a/action_plugins/launch_app/LaunchAppAction.qml
+++ b/action_plugins/launch_app/LaunchAppAction.qml
@@ -47,16 +47,16 @@ Item {
         }
 
         TextField {
-            id: _applicationToLaunch
+            id: _applicationPath
 
             Layout.fillWidth: true
 
             placeholderText: null != action ? null : "Enter an application to launch"
-            text: action.application_to_launch
+            text: action.application_path
             selectByMouse: true
 
-            onTextChanged: {
-                action.application_to_launch = text
+            onEditingFinished: {
+                action.application_path = text
             }
         }
 
@@ -71,7 +71,7 @@ Item {
         nameFilters: ["Application (*.bat, *.exe)", "All applications (*.*)"]
         title: "Select an application"
         onAccepted: {
-            _applicationToLaunch.text = selectedFile.toString().substring("file:///".length)
+            action.application_path = selectedFile.toString().substring("file:///".length)
         }
     }
 }
--- a/action_plugins/launch_app/LaunchAppAction.qml
+++ b/action_plugins/launch_app/LaunchAppAction.qml
@@ -1,0 +1,77 @@
+// -*- coding: utf-8; -*-
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Layouts
+import QtQuick.Window
+import QtQuick.Dialogs
+
+import QtQuick.Controls.Universal
+
+import Gremlin.Profile
+import Gremlin.ActionPlugins
+import "../../qml"
+
+
+Item {
+    property LaunchAppModel action
+
+    implicitHeight: _content.height
+
+    RowLayout {
+        id: _content
+
+        anchors.left: parent.left
+        anchors.right: parent.right
+
+        Label {
+            id: _label
+
+            Layout.preferredWidth: 150
+
+            text: "Application to launch"
+        }
+
+        TextField {
+            id: _applicationToLaunch
+
+            Layout.fillWidth: true
+
+            placeholderText: null != action ? null : "Enter an application to launch"
+            text: action.application_to_launch
+            selectByMouse: true
+
+            onTextChanged: {
+                action.application_to_launch = text
+            }
+        }
+
+        Button {
+            text: "Select application"
+            onClicked: _fileDialog.open()
+        }
+   }
+
+   FileDialog {
+        id: _fileDialog
+        nameFilters: ["Application (*.bat, *.exe)"]
+        title: "Select an application"
+        onAccepted: {
+            _applicationToLaunch.text = selectedFile.toString().substring("file:///".length)
+        }
+    }
+}
--- a/action_plugins/launch_app/__init__.py
+++ b/action_plugins/launch_app/__init__.py
@@ -1,0 +1,172 @@
+# -*- coding: utf-8; -*-
+
+# Copyright (C) 2024
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+from __future__ import annotations
+
+import logging
+from typing import List, TYPE_CHECKING
+from xml.etree import ElementTree
+
+from PySide6 import QtCore
+from PySide6.QtCore import Property, Signal
+
+from gremlin import event_handler, util
+from gremlin.base_classes import AbstractActionData, AbstractFunctor, \
+    Value
+
+from gremlin.profile import Library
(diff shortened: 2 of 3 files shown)


gremlin/ui/backend.py
```diff
@@ -276,6 +277,20 @@ def loadProfile(self, fpath):
         self.profileChanged.emit()
         signal.reloadUi.emit()
 
+    @Slot(str)
```
WhiteMagic: This should be in the functor not the backend singleton as that is really only intended for aspects the UI code itself has to communicate with.

gremlin/ui/backend.py
```diff
@@ -276,6 +277,20 @@ def loadProfile(self, fpath):
...
+
+        Args:
+            fpath: Path to the application being launched.
+        """
+
+        # ignore all exceptions from subprocess
+        try:
+            subprocess.run([fpath], check=False, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        except subprocess.SubprocessError:
+            pass
```
WhiteMagic: I'd log the failure as a warning to the log file so there is evidence something went wrong.

gremlin/ui/backend.py
```diff
@@ -276,6 +277,20 @@ def loadProfile(self, fpath):
...
+    def launchApp(self, fpath):
+        """Spawns a subprocess to launch an application
+
+        Args:
+            fpath: Path to the application being launched.
+        """
+
+        # ignore all exceptions from subprocess
+        try:
+            subprocess.run([fpath], check=False, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
```
WhiteMagic: Are these arguments necessary? The defaults should have the exact same behavior.

action_plugins/launch_app/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    ]
+
+    def __init__(
+            self,
+            behavior_type: InputType = InputType.JoystickButton
+    ):
+        super().__init__(behavior_type)
+
+        # Model variables
+        self.application_to_launch = ""
```
WhiteMagic: I probably would have called it `application_path` or `executable_path`. It also may be useful to be able to specify command line arguments. I think some people use such functionality with vendor specific joystick rgb control tools.

action_plugins/launch_app/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _from_xml(self, node: ElementTree.Element, library: Library) -> None:
+        self._id = util.read_action_id(node)
+        self.application_to_launch = util.read_property(
+            node, self.tag, PropertyType.String
```
WhiteMagic: Maybe cleaner to not reuse self.tag even if the XML tag will be the same as future changes might cause odd unwanted problems, like changing the tag but not wanting to update the XML one.

action_plugins/launch_app/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+        if not self.is_valid():
+            raise GremlinError(f"{self.application_to_launch} does not exists or is not accessible.")
```
WhiteMagic: If we throw an error for this here it should be a ProfileException. It's one of the very few specific ones that exists. Not sure we want to crash Gremlin for this problem though as the is_valid should result in it not being run and the functor will also bail on running it.

action_plugins/launch_app/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+            parent: QtCore.QObject
+    ):
+        super().__init__(data, binding_model, action_index, parent_index, parent)
+
+    def _qml_path_impl(self) -> str:
+        return "file:///" + QtCore.QFile(
+            "core_plugins:launch_app/LaunchAppAction.qml"
+        ).fileName()
+
+    def _action_behavior(self) -> str:
```
WhiteMagic, line 89: Just for your info, this is new code that I needed for the `hat_buttons` action as the action should only be added to hat inputs but the content of that action should treat the hat input as buttons such that it shows valid actions.

action_plugins/launch_app/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+        return self._binding_model.get_action_model_by_sidx(
+            self._parent_sequence_index.index
+        ).actionBehavior
+
+    def _get_application_to_launch(self) -> str:
+        return self._data.application_to_launch
+
+    def _set_application_to_launch(self, value: str) -> None:
```
WhiteMagic: Unsure the explicit conversion to string is needed, definitely won't hurt, but not sure that we could get a non string from anywhere and realistically if then Gremlin should scream.

action_plugins/launch_app/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+    def __call__(
+            self,
+            event: Event,
+            value: Value,
+            properties: list[ActionProperty]=[]
+    ) -> None:
+        if not self._should_execute(value):
+            return
+
+        logging.getLogger("system").debug(
```
WhiteMagic: Unsure I would log things like this, or if maybe should add an option to set the log level. Logging a failure to run though is absolutely good.

action_plugins/launch_app/__init__.py
```diff
@@ -0,0 +1,172 @@
...
+            f"Launching application ... {self.data.application_to_launch}"
+        )
+
+        be = backend.Backend()
```
WhiteMagic: As mentioned in the backend file, the logic should just live here as functors don't have any business interacting with that unless everything else is way too painful which it shouldn't since they live in python space.

action_plugins/launch_app/LaunchAppAction.qml
```diff
@@ -0,0 +1,77 @@
...
+
+        Label {
+            id: _label
+
+            Layout.preferredWidth: 150
+
+            text: "Application to launch"
+        }
+
+        TextField {
```
WhiteMagic, line 49: Question here is maybe if you want to allow the user to manually modify the executable or not. If not the text field can be made non-editable. If on the other hand they should edit it, maybe we can use a verification callback slot in the model with `file_exists_and_is_accessible` to ensure the final file is valid.

gremlin/ui/backend.py
leon-gh: OK.

gremlin/ui/backend.py
leon-gh: n/p.

action_plugins/launch_app/__init__.py
leon-gh: Renamed to application_path.

action_plugins/launch_app/__init__.py
leon-gh: n/p, I thought that changing tag once publishing is considered a no-no.

action_plugins/launch_app/__init__.py
leon-gh: n/p, changed to ProfileError. I think loading a bad profile could lead to unexpected behavior during use, so it's better to warn a user to fix their profile before activating it. Let me know how you want it done.

action_plugins/launch_app/__init__.py
leon-gh: I got inspired by other plugins, removed str()s.

action_plugins/launch_app/__init__.py
leon-gh: I thought debug was for that, n/p. Removed the log.

gremlin/ui/backend.py
leon-gh: Ahh, good catch. I just copied from load_profile, which needed to be in the backend. Moved it to functor.

action_plugins/launch_app/LaunchAppAction.qml
leon-zyprova: Updated TextField to trigger action updates only onEditingFinished and fixed FileDialog.onAccepted event to set action.application_path.
Also added throwing GremlinError in _set_application_path if file does not exist, but it shows in console only, not a pop-up as it would be nice.
Any suggestions?

leon-gh: I'd like to give the user ability to edit - makes it easier to copy-paste stuff, so you don't have to edit xml every time.
Let me know how you want it done and I'll modify it, if needed.

Updated TextField to trigger action updates only onEditingFinished and fixed FileDialog.onAccepted event to set action.application_path.
Also added throwing GremlinError in _set_application_path if file does not exist, but it shows in console only, not a pop-up as it would be nice.
Any suggestions?

WhiteMagic: That makes sense and I agree that it makes sense to let users modify by hand the content. I had hoped that `TextLField` had a good validation system, but the one it has https://doc.qt.io/qt-6/qml-qtquick-textinput.html#validator-prop is limited to a few presets that are not useful to us. So the way I'd do this is:
1. define a property like `isInputValid` in the model which checks if the file exists
2. change the outline based on that boolean, e.g. red if not valid and blue or nothing when it is valid
3. whenever the content changes the connections via signals should let QML know the value changed and thus update the UI

That way you don't have to throw an exception, which would be awkward while editing, but the user would still get a visual indication that something is wrong, since we're all trained to spot "red fields"

The whole exception and popup stuff used to work but it's misbehaving but I haven't looked into it and it may also only misbehave when run via the IDE but be ok when run as an application. Though I wouldn't inform the user about this issue with an exception as that'd get really annoying quickly. To me exceptions are really "something is brokenly wrong", like vJoy device missing etc.

leon-gh: I would not want to throw the exception either, but just some kind of pop-up or something to alert the user that something is wrong. I like red/blue outline, but I'm afraid users might not see it unless it's blinking or so, and it's so easy to "go to the next" button then realizing the old one didn't save your setting.

WhiteMagic: Imo, there is only so much helping one can do. If someone can't see a clearly marked error area, that's on them; I wouldn't want to subject competent users to annoying modal windows because of that. Exceptions in Gremlin are really for things that can't possibly work without being addressed, whereas here, it just won't execute a program.

Something down the road to have is a display/report thing that much more clearly shows/highlights issues with a configuration so that people have a way to see issues. Easiest instance of that is a warning sign on the inputs (again not all are visible all the time) and better is some side panel that shows error things like most IDEs do these days.

action_plugins/launch_app/__init__.py
WhiteMagic: Yeah that can work though may also lead to "broken" profiles that trip people in how to fix this, though that's something in general that needs to be looked at at some point, i.e. how to handle "bad" profiles.

WhiteMagic:
Gremlin got a "Run Command" action which does the same job as this, thus closing this PR.

PR #609, opened 2025-04-26
Add a framework for integration tests, and a simple test

code-monet:
Let me know if you think these will be useful, I can add some more tests in follow-up PRs. I hope this kind of automated testing will make development work faster and more enjoyable :)


WhiteMagic:
I think in general, a sort of hardware-in-the-loop testing can be useful though there are several things to consider.

1. Anything that's general action or other logic should be tested first as a separate module similar to a bunch of the existing tests
2. Mangling of profiles should as much as possible rely on Gremlin's code base rather than manual XML hacking as there is zero guarantee the XML structure will remain unchanged and it's also clearer what the intent is
3. Exploit the fact that Gremlin can fake physical device events without a real device. One can just create and emit events mimicking the profile without needing to abuse a vJoy device for input and output
4. The intermediate output system in combination with the above would allow to test complex interactions without relying on vjoy or physical devices to exist
5. This form of testing should focus on the interplay between actions
6. Given how complex profiles and tests might get it could be ideal to have a separate configuration that goes together with a profile that "describes" the kind of tests one wants to run together with the expected outputs. Maybe a sort of "scripted" playback traces or similar.

So in general this can be useful but one has to be careful with regards to how easy it is to create new tests and maintain the existing ones for this to be truly useful.

code-monet:
My thoughts:

1. Agreed that unit tests should also exist. However, integration tests can precede unit tests. My goal here is to contribute integration tests that need _less_ maintenance than unit tests, in part by minimizing test setup and fakes based on "internal" APIs.
2. Up to you - if you plan to keep the Gremlin profiles Python API stable, then I could switch the tests over. Otherwise, the code mangling the profiles is very straightforward. Using the Gremlin API also seemed like it would be a lot more lines of code.
3. Can you point me to this "faking" feature? But also, my next point:
4. Yes, but that's both less coverage than, and deviates from actual usage, compared to an integration test that works through vjoy and dill. I would instead add an (integration) test module with intermediate-output-based profiles to cover that feature.
5. Up to you! Integration tests can basically focus on anything you want. I'll add a framework in this PR, and some more test modules in later PRs. Feel free to list test cases you want covered and I'll take a stab at them.
   1. We could have a second directory of integration tests that are focused on that (and possibly bypass dill and vjoy), while this directory is more for general coverage. 
6.  I'm open to that, we can take a look when we write such tests. 

Agree that we should emphasize long-term maintainability and ease of adding new tests.

WhiteMagic:
I think to me it's not clear what these tests actually aim at verifying. Is it that the physical to virtual pathway, i.e. dill to vjoy, works? If so then then the entire aspect of profiles and actions is superfluous. If on the other hand the goal is to test the execution logic of actions, especially complex ones, then in reality you don't even need a complete Gremlin instance but only a `Profile` instance and a `CodeRunner` instance.

I don't think either way is wrong, but depending on what the goal is, the execution might need to look different to be easy to use and maintain. In my experience, most of Gremlin's bugs stem from the unexpected interplay of actions. The hardware handling within Gremlin never really had big bugs because it's actually quite simple. Dill and vJoy are the components that don't really have bugs. Dill doesn't like XInput and messed up DirectInput implementations, vJoy is finicky but neither of those are a thing Gremlin can fix. The main bugs Gremlin runs afoul are: a) UI fails to save state properly, b) actions interact in confusing ways, and c) logic errors in actions.

The fake event emission happens via the `EventListener` (https://github.com/WhiteMagic/JoystickGremlin/blob/3b734066530e9fed83a2a62acabb37df8424ccf6/gremlin/event_handler.py#L163) and an example of it being used can be found in the `macro` module (https://github.com/WhiteMagic/JoystickGremlin/blob/3b734066530e9fed83a2a62acabb37df8424ccf6/gremlin/macro.py#L417).

The point about XML mangling vs. action/profile API is that I would assume that reading the API methods makes it easier to understand what the goal of the test is without knowing how Gremlin stores this stuff internally. To mangle the XML one has to actually understand the structure and can't benefit from the tests that ensure actions do that part of parsing/saving correctly.

code-monet:
Re: goals, more the former, except even broader scope. This kind of end-to-end integration testing verifies that _all_ the components work together - (incomplete list I'm sure) Dill, vjoy, the PySide6 framework, the QMLs, Python, and indirectly, the MSVC runtime, and Windows features. 

For example, right now if you look at test/integration/test_e2e_profile_simple.py lines 63-64, there is a discrepancy for unknown reasons in the axis values. It's minor enough that it can be ignored, but it would potentially not be caught at all if the culprit is Dill/DirectInput or vJoy and we exclude it.

Similarly, say you upgrade to Python 3.13, and some [ctypes change](https://docs.python.org/3/whatsnew/3.13.html#ctypes) breaks the Dill or vJoy wrappings (this specific example might not, but you get the idea), but if you don't use the real Dill or vJoy in the E2E test, the regression won't be caught.

Finally, we might actually want to make changes to Dill and vJoy. For one, I hope to convince you to upgrade the vJoy version in Gremlin ;) Having such tests in place would give greater confidence.

So the goal would be: verify that common uses cases are working, end-to-end, using as many real dependencies as possible. This will be especially useful if we:

1.  Upgrade any of the direct or indirect dependencies
2.  Make changes to internal APIs that are mocked out in some unit tests
3.  Depend on external features, such as axis ranges or cpoints in DirectInput

---

One big downside is that these tests cannot practically be run in the GitHub Continuous Integration environment. Given what you said, it would make sense to add separate tests using fake physical events, intermediate outputs, and just the `CodeRunner` and `Profile` instances, focused on testing actions. These should probably be automated via GitHub Actions.

---

In terms of next steps, how about:

1. I'll modify this PR to use a `Profile` instance to modify the input profile, and then pass it to you for review?
2. In a follow-up PR, I'll fix the current broken unit tests and add them to CI
3. In a follow-up PR, I'll try add a (simpler) framework for the narrowly-scoped integration tests discussed above for testing the interplay of Actions. Add this to CI
4. I'll make some more PRs to improve coverage

WhiteMagic:
Ok that clarifies it a lot, thanks. In the beginning I wasn't sure whether the goal was complicated actions or more of the entire thing but since it's more of a complete system then we can ignore intricate stuff and see how to make testing the whole glued-up thing. And yes this definitely wouldn't be able to run on github runners, but that's not really an issue to me.

Since the main goal is to "simulate" interacting with Gremlin via "real" (at least to Windows) devices, it could be interesting to have that as a class or set of functions, that can easily be interfaced with to simplify future test cases. You seem to have done some aspects of that in the `conftest.py` file. I don't know how that would translate or be reusable if one wanted to test an aspect not requiring a profile.

Seeing how you look at dill's state to verify success made me think of another test that in this regime would be a good thing to have which is to ensure that what Gremlin's internal state belief is matches that of dill. The underlying assumption is that whatever dill sends Gremlin is also properly replicated which is needed for conditions and the like. That could be another use case to think of when thinking about the structure of the whole framework. I mainly mention this as I usually think of all the ways I may want to use something to figure out what the structure should look like.

Regarding the XML / profile mangling, depending on how complex of an interaction you want I think there might be an even simpler and more robust way. If all we need to use are vJoy remaps with the intent to verify values being passed around, surviving the round trip, then I would actually generate the profile programmatically. And similar in the test in I would have the values be swept or randomized with some obvious instances enforced to be checked like min/max/center. This would allow us to ensure we either check the full range or don't accidently picked some "this just happens to work" values. With that you could also get more creative and ensure things such as average delay is within expected bounds etc. (All that is more of a future thing but would be easy to add if the things were a bit programmatic too).

I would skip point 3 for now, at least until you're happy with the system-level testing part, as the action interplay would be a pretty separate bit from this more system-focused one. For 1, if we don't need complicated profiles, direct XML mangling works, using profile classes would likely be more readable, and what I described above might easier to do if you go with profiles anyways.

About the changing vJoy version. Per se I'm not opposed, it's just that from the commits and work I've seen on some of the forks I don't have a lot of faith in what has been done especially not given the issues they have caused in Gremlin. So that's why I'm very careful with telling people to use something other than what definitely works as vJoy is finicky enough as is.

code-monet:
Let me know of a test case that does not require a profile and I'll prototype it. We can also come back to this in the future, happy to refactor as needed to support more types of testing.

Re: checking Gremlin's internal state, that's a good idea! I'll add some commits when I have a chance to work on them. I imagine you're talking about `gremlin/ui/device.py` and perhaps `gremlin/input_cache.py`? I'll have to navigate the code a bit to see what other internal device states are being tracked.

Re: profiles, I did originally want to generate them programmatically, but got lured by the simplicity of the hack :) let me look into adding/switching to profile generation (and compare that with mangling the profile through the API). What I'd do is, generate the profile programmatically, write it to a file, and then let gremlin use it as it normally does - just to make sure that the profile loading path is also included in the whole end-to-end verification.

(It could also be worth using the generation code to verify/update some "golden profiles"; I'll take a look at the existing `test_profile.py`).

Re: inputs being swept - it's kinda like that right now with parameterized tests, I can definitely add more parameters. Randomizing inputs goes into the realm of fuzz testing which we can come back to after developing out this framework a bit more (so, agreed with you on it being a future thing). For one, we would want to add a way to verify that none of the inputs other than the intended ones are changing.

The changing-vJoy version is a future thing (say, post the initial R14 releases), and not necessary, as I can maintain a slight fork with the FFB features.

test/integration/conftest.py
```diff
@@ -0,0 +1,292 @@
...
+    override this fixture in test modules to use a generated profile.
+
+    (Where "real" means actual vJoy devices on this system.)
+    """
+    profile = gremlin.profile.Profile()
+    profile.from_xml(profile_path)
+    # Replace the (only) input device.
+    assert len(profile.inputs) == 1
+    _, input_items = profile.inputs.popitem()
+    profile.inputs[vjoy_di_device.device_guid.uuid] = input_items
```
WhiteMagic, line 93: Possibly can get away with `profile.inputs[vjoy_di_device.device_guid.uuid] = profile.inputs.pop()`

test/integration/conftest.py
```diff
@@ -0,0 +1,292 @@
...
+
+# +-------------------------------------------------------------------------
+# | Common fixtures, typically there should be no need to override.
+# +-------------------------------------------------------------------------
+
+
+@pytest.fixture(scope="module")
+def profile_path(profile_name: str) -> str:
+    """Returns profile path. Requires test modules to define the profile_name fixture."""
+    return os.path.join(os.path.dirname(__file__), "..", "xml", profile_name)
```
WhiteMagic: If we intend the integration XML profiles to be independent of the unit tests I'd probably have a XML folder in there. We then also should move the unit tests into a unit_test subfolder so the different purposes is clearly visible.

test/integration/conftest.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+@pytest.fixture(scope="module")
+def edited_profile_path(
+    profile_for_testing: gremlin.profile.Profile
+) -> Iterator[str]:
+    with tempfile.NamedTemporaryFile(delete_on_close=False) as f:
+      f.close()
```
WhiteMagic: Nitpicking: this looks like 2 rather than 4 spaces

test/integration/conftest.py
```diff
@@ -0,0 +1,292 @@
...
+
+    We return IDs instead of the control devices to not acquire vJoy devices a
+    test might not actually use.
+    """
+    # TODO: Share this functionality with joystick_handling.py
+    vjoy_ids = []
+    vjoy_proxy = gremlin.joystick_handling.VJoyProxy()
+    for i in range(1, 17):
+        if vjoy.device_exists(i):
+            try:
```
WhiteMagic, line 154: I suspect here having `vjoy_ids.append(i)` in the try block  would result in the same behavior though I'm also fine with the slightly more explicit messaging of the `else` path.

test/integration/test_e2e_profile_simple.py
```diff
@@ -0,0 +1,129 @@
...
+        vjoy_output: int,
+    ):
+        input_axis_id = 1
+        output_axis_id = 3
+        vjoy_control_device.axis(linear_index=input_axis_id).set_absolute_value(
+            di_input
+        )
+        # There is a small discrepancy between these values, for unknown reasons
+        # (unclear if this is a Gremlin bug).
+        app_tester.assert_axis_eventually_equals(
```
WhiteMagic: I just checked in DILL if there is anything odd with axis value reading but all it does is grab the information out of the struct and send it off to the callback. And the datatype is a long on both sides so it shouldn't do odd things with that. Maybe there is some odd calibration thing going on in Windows itself? Thought might be worth checking what other programs report numerically.

test/integration/test_e2e_profile_simple.py
```diff
@@ -0,0 +1,129 @@
...
+        vjoy_control_device.button(index=input_button_id).is_pressed = di_input
+        app_tester.assert_button_eventually_equals(
+            vjoy_di_device, input_button_id, vjoy_output
+        )
+        app_tester.assert_button_eventually_equals(
+            vjoy_di_device, output_button_id, vjoy_output
+        )
+
+    @pytest.mark.parametrize(
+        ("di_input", "vjoy_output"),
```
WhiteMagic: There is a fun bug that might occur with hats. Dill forwards the actual integer value, i.e. doesn't make sure it's one of the 9 "valid" options. So I'm not positive Gremlin won't throw a crazy fit if for some reason it got a different value, though that's not testable via the profile aspect.

WhiteMagic:
> Let me know of a test case that does not require a profile and I'll prototype it. We can also come back to this in the future, happy to refactor as needed to support more types of testing.

I think the internal state one is probably the best candidate honestly.

> Re: checking Gremlin's internal state, that's a good idea! I'll add some commits when I have a chance to work on them. I imagine you're talking about `gremlin/ui/device.py` and perhaps `gremlin/input_cache.py`? I'll have to navigate the code a bit to see what other internal device states are being tracked.

Yeah though I've just munched those two into one file. The other file to look at is the `event_handler.py` as that one feeds these the storage classes as they bypass the QT event system to ensure that the state of the buffers is updated before actions can execute. This is in part to avoid the fun issue in R13 where people have a condition on the triggering input and wonder that nothing works as expected, as the buffer (which the condition uses) hasn't been updated.

> Re: profiles, I did originally want to generate them programmatically, but got lured by the simplicity of the hack :) let me look into adding/switching to profile generation (and compare that with mangling the profile through the API). What I'd do is, generate the profile programmatically, write it to a file, and then let gremlin use it as it normally does - just to make sure that the profile loading path is also included in the whole end-to-end verification.

If we want to test the profile loading in the whole thing as well that makes sense. If we delegate that to another testing part then obviously it wouldn't be that necessary here. Mostly also if it makes things easier.

> (It could also be worth using the generation code to verify/update some "golden profiles"; I'll take a look at the existing `test_profile.py`).

That one is not necessarily the most complete profile as it was done a long long time ago. Importantly it won't have awkward actions such as merge axis in it.

> Re: inputs being swept - it's kinda like that right now with parameterized tests, I can definitely add more parameters. Randomizing inputs goes into the realm of fuzz testing which we can come back to after developing out this framework a bit more (so, agreed with you on it being a future thing). For one, we would want to add a way to verify that none of the inputs other than the intended ones are changing.

Makes sense and both sorts of tests have their place I would say.

> The changing-vJoy version is a future thing (say, post the initial R14 releases), and not necessary, as I can maintain a slight fork with the FFB features.

Sounds good.

test/integration/conftest.py
code-monet: `dict.pop` requires a key, which we don't have (or need); `popitem` is the cleanest way in this case.

test/integration/conftest.py
code-monet: I'll keep the more explicit version, not much more code and a bit clearer as to which lines might actually raise.

test/integration/conftest.py
code-monet: Indeed! Fixed.

test/integration/conftest.py
code-monet: Agreed, done.

test/integration/test_e2e_profile_simple.py
code-monet: The deltas are quite large (defined in `conftest.py` now), and doesn't seem to be affected by me doing a "reset to default" from the Windows USB Game Controllers menu. I can look into this more later.

test/integration/test_e2e_profile_simple.py
code-monet: If I understood you correctly... I have implemented an additional test case showing current behavior (Gremlin leaves the hat at Center).

code-monet:
Added assertion methods for the `input_cache`. All state storages are covered now, right? The state stored in Qt is being verified via what is eventually written to the output device, and the `input_cache` checks were just added.

> If we want to test the profile loading in the whole thing as well that makes sense. If we delegate that to another testing part then obviously it wouldn't be that necessary here. Mostly also if it makes things easier.

Understood - writing the profile to a file and then having Gremlin use that was just a few lines of code, so I'll leave it like that. The fixture for getting the Python `Profile` instance is also available if we wanted to bypass the file write/read in other tests.

test/integration/test_e2e_profile_simple.py
WhiteMagic: Ok then at least it behaves gracefully (or other aspects in the whole pipeline correct it for it) But good to have it in there as a "just in case".

WhiteMagic:
This looks good to me overall, let me know when you went over it again, as I likely broke quite a few things with the reshuffling of files and contents due to the cyclical import issues.

test/integration/test_e2e_profile_simple.py
code-monet: Yeah would be good to know if it starts crashing one day (and also good to have in case we need to be able to handle such inputs one day).

code-monet:
Luckily (or is it the magic of E2E testing? :) I only had to change the imports. The integration tests are running and passing, and I'll send a PR to fix the unit tests.

Please go ahead and merge this PR.

WhiteMagic:
Done, thanks for all the work on this.

I also had a quick look at the delta required for the tests to pass and it's exactly the same on my system. Though the delta isn't the same for all axes which is even weirder.

code-monet:
Happy to help! And a perpetual thanks to you for starting and maintaining this project.

I started looking into the axis delta situation - I didn't realize that it's not the same for all axes, that's extra weird, thanks for sharing that. I'll keep digging and send PRs your way as I find stuff.

WhiteMagic:
I don't know about different axes but what I saw was that for some values, negative ones specifically, a larger delta was required for them to pass than for other values. I didn't check if the order in which the test uses the values matters. It's almost as if there is some curve or similar somewhere.

Issue #613, opened 2025-05-17
Fixing broken unit tests


code-monet:
Opening this mainly for tracking that I'll send a [PR equivalent to this one](https://github.com/code-monet/JoystickGremlin/pull/1) once #609 is merged (as it depends on that).

code-monet:
These are *almost* fixed up. There are still a couple of issues:

1. The unit tests won't run if the profile directory doesn't already exist.
2. Some of the unit tests require a vJoy device to be present on the machine (shown below)

I'm trying to address these [here](https://github.com/code-monet/JoystickGremlin/pull/3), but a bit more work is needed. Meanwhile, can you let me know if you would be interested in having such a CI? Or maybe you prefer just to run the tests locally?

---

Unit tests needing a real vJoy device (I can probably work around this by patching in a made-up vJoy device for the unit tests)

```
FAILED test/unit/test_action_map_to_vjoy.py::test_ctor - IndexError: list index out of range
FAILED test/unit/test_action_map_to_vjoy.py::test_actions - IndexError: list index out of range
FAILED test/unit/test_action_map_to_vjoy.py::test_from_xml - IndexError: list index out of range
FAILED test/unit/test_action_map_to_vjoy.py::test_to_xml - IndexError: list index out of range
FAILED test/unit/test_modes.py::TestModeHierarchy::test_complex_modifications - gremlin.error.ProfileError: "Unknown type 'map-to-vjoy' in action with id '837322a1-9c25-42d1-ad1a-d134e692885a'"
FAILED test/unit/test_profile.py::test_simple_action - gremlin.error.ProfileError: "Unknown type 'map-to-vjoy' in action with id 'd67cbad2-da3f-4b59-b434-2d493e7e6185'"
```

WhiteMagic:
I think having tests in a CI won't hurt as that way things won't silently be overlooked (assuming one checks the CI status :-)).

With regards to the tests requiring a real vJoy device. One option is as you suggest mocking that part, though the other option is also to detect that and skip those tests.

code-monet:
Hehe, sounds good. I considered a few options, including skipping, but I figured adding fake devices could come in handy later for additional unit testing.

PR #614, opened 2025-05-20
Fix all tests so that we can just run "pytest"

code-monet:
A few things were going on:

1. The config test was mutating the config singleton, removing the actions etc (which normally are added during module import), and this broke many of the other unit tests. The fix is to have config test use its own instance of a Configuration rather than the singleton.
2. The integration test imports `joystick_gremlin`, which changes the working directory. The unit tests run afterwards and can't find the XMLs anymore. Fix by using module-relative paths.
3. Dill can only be initialized once. To allow tests to be run individually as well as together, the `DILL.init()` isn't removed from the unit tests' `conftest.py`, the DILL class instead only allows a single `init()` during process lifetime.
4. Reduce scope of package fixtures to `package` rather than `session`


code-monet:
Fixes #613

gremlin/config.py
```diff
@@ -65,10 +65,13 @@ def count(self) -> int:
         """
         return len(self._data)
 
+    def _skip_reload(self) -> bool:
```
WhiteMagic: I personally would name this `_should_skip_reload` or similar. I tend to name properties of methods that return a boolean like a "question" or similar like `is_valid` so it's clearly visible to me that it will tell me yes / no. Having it be `_skip_reload` I wouldn't be certain if I mean "this has skipped reloading" or "I don't need to relaod this".

Also what was the rational for extracting the single check that from memory is only used there into a standalone method?

test/unit/conftest.py
```diff
@@ -4,6 +4,8 @@
 
 sys.path.append(".")
 
+import os.path
```
WhiteMagic: If you prefer you can also use `pathlib.Path` for all the path mangling which is what I try to use in most new Gremlin code I write. But both are fine in my eyes.

code-monet: Switched over.

gremlin/config.py
code-monet: Renamed. This function is patched out in test_config, otherwise the load() call in the test would be no-op if not enough time had passed since the last load (leading to different behavior when running the test by itself, vs when running with all the other unit tests).

code-monet:
Changes implemented, please go ahead with merge if not more requests.

WhiteMagic:
Thanks for fixing the tests up.

PR #615, opened 2025-05-22
Add ci workflow to run unit tests

### LINKED ISSUE #613: Fixing broken unit tests

code-monet:
I had to modify a bit of Gremlin internals to make all this work. Specifically:

1302efd03a1a72d76fe1bb061694add8456ba529: changes how the `Configuration` file watcher starts. Need to prevent access violation on exit from unit tests (integration tests don't have this issue because they run/initialize(?) the Qt app - my comment for `ensure_watcher_running` might be inaccurate).

54b66c1fcc18c0be0e1315667422f31c779da053: Does not throw an exception if the configuration file was not found when attempting to save it. The idea is that this shouldn't happen during normal Gremlin execution because the profile directory is created at the module level in `joystick_gremlin.py`. Let me know if you have a better idea.

Fixes #613

### DIFF
--- a/gremlin/config.py
+++ b/gremlin/config.py
@@ -54,8 +54,13 @@ class Configuration:
         self._last_reload = None
         self.load()
 
-        self.watcher = QtCore.QFileSystemWatcher([_config_file_path])
-        self.watcher.fileChanged.connect(self.load)
+        self.watcher = None  # See ensure_watcher_running()
+        
+    def ensure_watcher_running(self):
+        """Only call this after configuring the rest of the Qt app."""
+        if self.watcher is None:
+            self.watcher = QtCore.QFileSystemWatcher([_config_file_path])
+            self.watcher.fileChanged.connect(self.load)
 
     def count(self) -> int:
         """Returns the number of parameters stored.
--- a/gremlin/config.py
+++ b/gremlin/config.py
@@ -54,14 +54,6 @@ class Configuration:
         self._last_reload = None
         self.load()
 
-        self.watcher = None  # See ensure_watcher_running()
-        
-    def ensure_watcher_running(self):
-        """Only call this after configuring the rest of the Qt app."""
-        if self.watcher is None:
-            self.watcher = QtCore.QFileSystemWatcher([_config_file_path])
-            self.watcher.fileChanged.connect(self.load)
-
     def count(self) -> int:
         """Returns the number of parameters stored.
 
@@ -138,15 +130,13 @@ class Configuration:
                 "expose": entry["expose"]
             }
 
-        try:
-            # Write data to file
-            with open(_config_file_path, "w") as hdl:
-                encoder = json.JSONEncoder(sort_keys=True, indent=4)
-                hdl.write(encoder.encode(json_data))
-        except FileNotFoundError as e:
-            logging.getLogger("system").exception(
-                f"Failed to save configuration file: {e}"
+        # Write data to file
+        with open(_config_file_path, "w") as hdl:
+            encoder = json.JSONEncoder(
+                sort_keys=True,
+                indent=4
             )
+            hdl.write(encoder.encode(json_data))
 
     def register(
         self,
--- a/gremlin/config.py
+++ b/gremlin/config.py
@@ -133,13 +133,15 @@ class Configuration:
                 "expose": entry["expose"]
             }
 
-        # Write data to file
-        with open(_config_file_path, "w") as hdl:
-            encoder = json.JSONEncoder(
-                sort_keys=True,
-                indent=4
+        try:
+            # Write data to file
+            with open(_config_file_path, "w") as hdl:
+                encoder = json.JSONEncoder(sort_keys=True, indent=4)
+                hdl.write(encoder.encode(json_data))
+        except FileNotFoundError as e:
+            logging.getLogger("system").exception(
+                f"Failed to save configuration file: {e}"
             )
-            hdl.write(encoder.encode(json_data))
 
     def register(
         self,
--- a/joystick_gremlin.py
+++ b/joystick_gremlin.py
@@ -327,6 +327,7 @@ def make_gremlin_app(argv):
     # Run UI
     syslog.info("Gremlin UI launching")
     app.aboutToQuit.connect(shutdown_cleanup)
+    cfg.ensure_watcher_running()
     return app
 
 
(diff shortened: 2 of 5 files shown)


code-monet:
Example workflow run: https://github.com/code-monet/JoystickGremlin/actions/runs/15177650097

gremlin/config.py
```diff
@@ -133,13 +138,15 @@ def save(self):
                 "expose": entry["expose"]
             }
 
-        # Write data to file
-        with open(_config_file_path, "w") as hdl:
-            encoder = json.JSONEncoder(
-                sort_keys=True,
-                indent=4
+        try:
```
WhiteMagic: This will "silently" keep failing, correct? Maybe here it could be good to attempt to create the file and if that doesn't work actually throw an exception to terminate as that would indicate something is truly wrong.

test/unit/conftest.py
```diff
@@ -2,19 +2,104 @@
...
+    """Creates a repeatable, faked DeviceSummary."""
+    # Data below was generated from a vJoy device.
+    guid = dill._GUID(
+        Data1=501018480 + int(is_virtual),
+        Data2=264,
+        Data3=4592,
+        Data4=(128, 4, 68, 69, 83, 84, 0, 0),
+    )
+    axis_map_array = (dill._AxisMap * 8)()
+    for i in range(8):
```
WhiteMagic: If you want to test a more complicated scenario then you can modify the axismap to say only have 6 axes and skip some in between. This in the past caused issues with calibration and other things as in places the mapping from index to axis name was not done correctly.

joystick_gremlin.py
```diff
@@ -327,6 +327,7 @@ def make_gremlin_app(argv):
     # Run UI
     syslog.info("Gremlin UI launching")
     app.aboutToQuit.connect(shutdown_cleanup)
+    cfg.ensure_watcher_running()
```
WhiteMagic: I would put that further up with the other config related calls, like `cfg.purge_unused()` around line 216.

WhiteMagic:
Regarding the file watcher, I suspect it requires the `QApplication` instance to exist in order to be created. There is actually also a custom `FileWatcher` class in `gremlin.utils`. I can't say why I haven't used that in the new config class, likely forgot I had written that one. That one should not depend on the QApplication instance to exist, so we might as well just swap over to that one.

The file writing is a bit more tricky. This should only really fail if the folder in which Gremlin tries to write doesn't exist to being with if my memory of `open`'s behavior is correct. Is the issue that it should throw an exception when it can't save or the fact that it fails? If it fails we maybe should find a way to convey where to store the file as in CI systems there might not be a %userprofile% folder or similar.

gremlin/config.py
code-monet: I can/should actually revert this change for now - the issue is addressed by importing `joystick_gremlin` before `event_handler` in `unit\conftest.py`.

Correct, this will fail a few times until the profile directory is created. We can probably do a better fix in a separate PR, I'm open to ideas (should be able to leverage my [unpaid intern](https://blog.google/technology/google-labs/jules/) for some types of large refactorings).

code-monet:
I'll take a stab at switching to your custom `FileWatcher` in a separate PR, just in case that snowballs into a bigger change.

I should have elaborated more regarding the file writing - it was throwing an exception on CI because the folder creation is initiated by the `joystick_import.py` module at the module scope, which wasn't happening if you just run `test\unit`. For now, I'm just importing that module in the right order in `unit\conftest.py`. (`integration\conftest.py` didn't have this problem).

1. I contemplated moving the folder creation into the `Configuration.save` function but that is a bit more magic than I like
2. We could instead have the `load` triggered `save` be "best effort" and have a direct `save()` call at the right place in `joystick_gremlin.py`
3. We could do a larger refactoring that makes the registration of actions etc. into `Configuration` via some explicit mechanism rather than at module import. Open to ideas and throwing the task at Jules bot!

code-monet:
Actually, when is the file watching expected to be useful? Looking at the code, the only write access to that file is from `save`. Is the idea that user plugins or a human could modify that file?

gremlin/config.py
WhiteMagic: I think if the import reorder fixes it for the test then that's probably ideal. To fix the issue in the config class itself probably could be as simple as calling the util function that creates the folder and files in there too, rather than just the `joystick_gremlin.py`, in case this happens somewhere else. Though that would not fix the folder not existing in CI environments etc. Though then that could be a case that the utility function should realize such that the outcome of running that function is always a "good state".

WhiteMagic:
> Actually, when is the file watching expected to be useful? Looking at the code, the only write access to that file is from `save`. Is the idea that user plugins or a human could modify that file?

Good question, in my current thinking there is no reason to interface with the config in any other way than that class, and modifying it manually isn't something a user should do. And if they do then that's a case where they should know what they're doing and thus also relaunch Gremlin. The reloading was added for the old config class really early on and it could be that this was such that transformations of the configuration data would be applied upon saving a modified option. This isn't happening any longer with the dynamic config class as really just a fancy key-value store and no special handling. So I'm fine with us dropping the whole file watching entirely.

gremlin/config.py
code-monet: Reverted to original.

Yeah we could ensure that the folder exists during save, if this becomes an issue again. The CI environment does have a user profile directory, and the gremlin profile directory does eventually get created, so we're good for this PR.

joystick_gremlin.py
code-monet: Watcher removed per discussion in the main thread.

test/unit/conftest.py
code-monet: Done, used a 4 axis controller for reference and added a few more axis at the end.

code-monet:
Changes made, please take a look.

[CI run with latest code](https://github.com/code-monet/JoystickGremlin/actions/runs/15228791545/job/42834100114)

WhiteMagic:
Awesome thanks, I'll see if the workflow does work on its own or needs poking. I'm also likely going to be a bit less responsive as I'm travelling most of June, though still will look at things.

code-monet:
Thanks for the heads-up! Any PRs I send, no rush.

PR #616, opened 2025-05-31
Improve precision in axis values

code-monet:
The following causes were found (and some were fixed, others just have comments):

1. Typo `32678 -> 32768`
2. In vjoy, float axis to integer conversion now uses rounding instead of truncation and does this only once
3. Add comment explaining remaining delta (not really worth fixing IMO)


pyproject.toml
```diff
@@ -15,6 +15,7 @@ PySide6 = ">=6.8"
 PyInstaller = ">=6.10"
 pywin32 = ">=308"
 pytest = ">=8.3"
+pytest-subtests = "^0.10.0"
```
WhiteMagic, line 18: I assume this is to provide more documentability to tests?

test/integration/app_tester.py
```diff
@@ -31,13 +31,19 @@
 _ASSERT_EVENTUALLY_MAX_DELAY = 1  # Seconds
 _ASSERT_EVENTUALLY_RETRY_DELAY = 0.01  # Seconds
 
-# These limits should be tighter, but the results failures need to be investigated.
-_INTEGER_AXIS_MAX_DELTA = 68  # Should be 1.
-_FLOAT_AXIS_MAX_DELTA = 0.004  # Should be 1 / 32767
+# The max delta below for integration testing is determined by the following factors:
+# 1. vJoy range is 0-32767, but DirectInput default axis range is 0-65535.
+# 2. The "calibration" of axis values for Dill reads is done on a different range than
```
WhiteMagic, line 36: What exactly is the difference between them? Modifying that in dill isn't an issue (assuming vJoy's behavior is correct).

vjoy/vjoy.py
```diff
@@ -189,7 +193,7 @@ def __init__(self, vjoy_dev: VJoy, axis_id: int) -> None:
             ctypes.byref(tmp)
         )
         self._max_value = tmp.value
-        self._half_range = int(self._max_value / 2)
+        self._half_range = (self._max_value - self._min_value) / 2
```
WhiteMagic, line 192: Is the min value actually not 0? The "off by 0.5" which is incurred by clamping the value is adding up the the error you see in the tests (after fixing the wrong range specification?)

vjoy/vjoy.py
```diff
@@ -273,7 +277,7 @@ def value(self, value: float) -> None:
         )
 
         if not VJoyInterface.SetAxis(
-                int(self._half_range + self._half_range * self._value),
+                _simple_round(self._half_range + self._half_range * self._value),
```
WhiteMagic: I would prefer this not to be a function call. They have some cost, and this is a function that is being called a lot so one of the cases where the overhead at some point could be measurable.

pyproject.toml
code-monet: It gives a bit more output, and more importantly, an assertion failing in a subtest will not prevent subsequent assertions from being run.

vjoy/vjoy.py
code-monet: I'm not sure I follow the question re: 0. Yes the min value is 0. Previously, the clamping would cause an off-by-1 error for zero and positive values (because the mid point and range were both clamped downwards, resulting in a total loss of 1).

vjoy/vjoy.py
code-monet: Sure, will inline this logic.

test/integration/app_tester.py
code-monet: The most direct way to avoid this issue would be to change vJoy to have the axis ranges be 0-65535 (I think I could do this by only changing vjoyconfig.exe and no need to change the driver - but we'd have to ship our own version of the app with Gremlin).

Other than that, we could try one of the following:

1. Change DirectInput axis ranges in Dill to -16384-16383.
2. Don't change DirectInput axis ranges at all; calibrate and pick midpoints ourselves.

The root of the problem is that our logic to convert values from `[-32768, 32767]` to `[-16384, 16383]`, slightly differs from how DirectInput seems to be doing it. For us, the step sizes for calibrated values don't exactly line up, seen below:

| Dill | vjoy |
| ----------- | ---- |
| 2 / 32767 ~=  6.1037e-05      |  1/ 16383 ~= 6.1039e-05 |

So calibrated floats for odd Dill values always round downwards when converting to vjoy "raw". But because of DirectInput special logic, the "What we get from Dill if we write x from vjoy" table instead looks like:

| Vjoy write | Dill read |
| -- | -- |
| 0 | -32768 |
| 1 | -32766 |
| 16383 | **-2** |
| 16384 | 0 |
| 16385 | **1** |
| 16386 | **3** |

So say we want to write "Dill 2" to vjoy, we calibrate that to [-1 ,1], uncalibrate in vjoy,py, and write 16385 to vjoy, but Dill instead reads back 1. And then when we write back "Dill 1" to vjoy, we instead get Dill readback of 0.

WhiteMagic: Ok I see it's fun with resolution changes being funky. Though with vJoy being 15 bit and DirectInput (and thus dill) being 16 bit there isn't much one can do I suspect. The 16 bit world will always have "unused" steps I would assume. The behavior around 0 is a bit odd, but that should be fine in the big scheme of things. No one will be able to tell the difference between 0 -> 1 and 0 -> 2.

For now I think this is more than good enough. If we wanted to fix this I would prefer Gremlin to be the place where things get fixed. I definitely don't want to mangle with vJoy itself and while dill could report 15 bit I don't see why it should drop resolution for every device if it's just vJoy that's being awkward.

code-monet: Agreed, the remaining delta isn't big enough to be a problem.

PR #617, opened 2025-05-31
Add workflow_dispatch to python-test.yml

code-monet:
Presumably this will allow repo maintainers to manually trigger checks on a PR.


WhiteMagic:
It should as I've used that on the pylint workflow. Not going to be used often I suspect since the tests will run on push and pull requests anyhow. Not sure the `:` is required or causes issues in the yaml, but easy to fix if it does.

PR #623, opened 2025-08-22
Update unit test now that types are preserved for list elements

code-monet:
CI should be happy with this commit.


WhiteMagic:
Yeah that was getting annoying for thing for the options dialogue, so had to find a way to avoid having to stringify lists for no good reason :-)

PR #624, opened 2025-08-25
Fix response curve action test and add a new test case demonstrating how to modify action parameters in the test

code-monet:
I will add more tests to this PR in the next few days to cover the other types of curves.


WhiteMagic:
Sounds good, I should also be done with the sweeping changes caused by renaming the intermediate output system.

code-monet:
Added the remaining test cases, please take a look.

I think I found a bug - the "symmetry" option for the cubic splines does not work, and I had to explicitly add points on both sides. In the UI, if I enable symmetry and add a point, the symmetric point is not added. I can fix that in a follow-up PR if you like.

code-monet:
Okay, added a commit for that as well.

I had to change some expected values very slightly for the cubic Bezier spline if I didn't add (0, 0) explicitly after enabling symmetry - not sure if it matters and it might actually be expected?

gremlin/spline.py
```diff
@@ -71,11 +69,29 @@ def add_control_point(self, x: float, y: float) -> None:
         the symmetric mode.
 
         Args:
-            x: x-coordinate of the control point
-            y: y-coordinate of the control point
+            x: x-coordinate of the control point (will be clamped).
```
WhiteMagic, line 72: One could think of a `RangeValue` data type that enforces this and also makes the behavior/intent clear as this sort of "value has to be within" comes up in a few places.

gremlin/spline.py
```diff
@@ -71,11 +69,29 @@ def add_control_point(self, x: float, y: float) -> None:
...
+            y: y-coordinate of the control point (will be clamped).
+            idx: index of the control point to set
+        """
+        x = util.clamp_analog_axis(x)
+        y = util.clamp_analog_axis(y)
+        points = self.control_points()
+        points[idx].x = x
+        points[idx].y = y
+        if self.is_symmetric:
```
WhiteMagic, line 90: The code didn't have this before either, but it may be worthwhile adding a check for the implied invariant that there is an odd number of control points when the symmetry mode is active.

WhiteMagic:
Looks good to me so far. I think there is also something horribly broken with the UI for the response curve, at least it misbehaved quite badly when I quickly tried it before. So another thing to revisit.

To me this looks fine and I can go ahead and merge it unless you have more tweaks you want to make to it.

code-monet:
Please go ahead and merge it if the latest commits look fine, thanks!

I didn't find the UI too broken, although I did notice that you can create spline curves that go out of bounds (didn't try running it):

image

gremlin/spline.py
code-monet: Added check and added a test case that uses this function.

gremlin/spline.py
code-monet: Here's [one way to do it](https://github.com/code-monet/JoystickGremlin/pull/4), let me know what you think and I'll open a PR

code-monet:
Hang on, not sure why the test is failing - investigating.

code-monet:
Okay, looks like adding that assert exposed a bug. Added a commit to fix the bug, let me know if more changes are desired, otherwise please go ahead with merging this PR.

WhiteMagic:
Looks all good to me. And the curves going outside the valid range is also in R13 so not a regression. I think there is code later in the pipeline that would result in those curve values being clamped if one attempted to send them to vJoy (but not 100% certain).

PR #627, opened 2025-08-30
Merge axis actiontest

code-monet:
As well as refactoring to share some fixtures between action tests.

### DIFF
--- a/test/integration/action_plugins/conftest.py
+++ b/test/integration/action_plugins/conftest.py
@@ -21,12 +21,8 @@ import pytest
 
 from action_plugins import map_to_logical_device
 import dill
-from gremlin import logical_device
-from gremlin import plugin_manager
-from gremlin import profile
-from gremlin import shared_state
-from gremlin.ui import backend
-from gremlin.ui import device
+from gremlin import logical_device, plugin_manager, profile, shared_state
+from gremlin.ui import backend, device
 
 LogicalActionCallableT = Callable[
     [logical_device.LogicalDevice.Input.Identifier | str],
@@ -41,9 +37,9 @@ LogicalIdentifierCallableT = Callable[
 def profile_for_test() -> profile.Profile:
     # This is important for locally-run tests where the last used profile could
     # get loaded, if present in configuration.
-    backend.Backend().profile = pr = profile.Profile()
-    shared_state.current_profile = pr
-    return pr
+    backend_ = backend.Backend()
+    backend_.newProfile()
+    return backend_.profile
 
 
 @pytest.fixture(scope="module")
--- a/test/integration/action_plugins/conftest.py
+++ b/test/integration/action_plugins/conftest.py
@@ -15,11 +15,90 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 
 import pytest
 
+from action_plugins import map_to_logical_device
+import dill
+from gremlin import logical_device
+from gremlin import plugin_manager
+from gremlin import profile
+from gremlin import shared_state
 from gremlin.ui import backend
+from gremlin.ui import device
+
+LogicalActionCallableT = Callable[
+    [logical_device.LogicalDevice.Input.Identifier | str],
+    map_to_logical_device.MapToLogicalDeviceData,
+]
+LogicalIdentifierCallableT = Callable[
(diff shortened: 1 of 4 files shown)


test/integration/action_plugins/conftest.py
```diff
@@ -15,11 +15,90 @@
...
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from collections.abc import Iterator
+from collections.abc import Callable, Iterator
 
 import pytest
 
+from action_plugins import map_to_logical_device
+import dill
+from gremlin import logical_device
```
WhiteMagic: I usually do something like `from gremlin import A,  B, C` when there are multiple imports from `gremlin` or `gremlin.ui`. Though this is also fine.

test/integration/action_plugins/conftest.py
```diff
@@ -15,11 +15,90 @@
...
+LogicalIdentifierCallableT = Callable[
+    [logical_device.LogicalDevice.Input.Identifier | str], device.InputIdentifier
+]
+
+
+@pytest.fixture(scope="module")
+def profile_for_test() -> profile.Profile:
+    # This is important for locally-run tests where the last used profile could
+    # get loaded, if present in configuration.
+    backend.Backend().profile = pr = profile.Profile()
```
WhiteMagic: If you use `Backend().newProfile()` all this should be performed and you don't run the risk of something like `shared_state` changing or being added, causing tests to break because the logic in here hasn't been updated. I would assume calling that method should be fine as you have to have the whole Qt machinery up anyways to create the `Backend` instance?

test/integration/action_plugins/test_merge_axis.py
```diff
@@ -0,0 +1,286 @@
...
+_LOGICAL_INPUT_AXIS1_LABEL = "InputAxis1"
+_LOGICAL_INPUT_AXIS2_LABEL = "InputAxis2"
+_LOGICAL_OUTPUT_AXIS_LABEL = "OutputAxis"
+
+
+@pytest.fixture(scope="module")
+def merge_axis_action() -> merge_axis.MergeAxisData:
+    """Create merge axis action instance for testing.
+    
+    Merge axis should be modifiable while Gremlin is running, so making
```
WhiteMagic, line 38: Just as an FYI this "live modification" feature is a happy accident in how I started using the data and functor classes. In R13 I copied the data into the functor while in R14 I decided that this was dumb and so just keep a reference to the data instance. The implication of this is that changes to the data class from the UI will be seen by the functor as they access the same instance. However, not every modification from the UI makes it into being updated live.

I haven't worked out where the boundaries are but one obvious case is that new actions, effectively modifications to the `CodeRunner` and `EventListener` lookup tables, do not get live updated. Also some kind of structural changes to an action doesn't fully makes it in the live updating. So there is definitely room for improvement. Though as I said it's more of a happy accident than a planned feature..

test/integration/action_plugins/conftest.py
code-monet: Reordered in next commit and noted for future ones!

test/integration/action_plugins/conftest.py
code-monet: Checked that it works, thanks! Doing this way in next commit.

test/integration/action_plugins/test_merge_axis.py
code-monet: Good to know! Ideally I wouldn't depend on non-features in the tests, but there is a definite performance benefit here (vs reactivating Gremlin for each test cases that modifies the action, but that's not a huge deal either). However I'm on board with this happy accident, as an end user it's nice to be able to tweak actions without toggling Gremlin to try them out. So I'll keep using this non-feature, and if I run into a situation where it isn't working, we'll fix it :)

code-monet:
Please go ahead with merge or let me know if more changes are desired.

code-monet:
(Rebased on top of dev)

test/integration/action_plugins/test_merge_axis.py
WhiteMagic: Oh yeah definitely it's something that a lot of people always wanted in R13 and R14 does it by accident to some extent. Will be some work to make it coherent in a smart way (the dumb way is to refresh the entire lookup table which may still be imperceptibly fast).

WhiteMagic:
Looks all good to me, and thanks a lot for the continued effort in adding test coverage to Gremlin.

code-monet:
Happy to help move this project closer to launch :)

PR #628, opened 2025-08-30
Add unit tests for user scripts module

code-monet:
One thing I noticed is that the virtual input variable seems to be incomplete - in the GUI you can only pick axes, not buttons or hats?

For the next couple of pull requests the plan is to add:

1. An end-to-end integration test using a vJoy device.
2. An integration test using logical devices, if these are available in user scripts (or the idea is to eventually have them available there?)

### DIFF
--- a/test/unit/conftest.py
+++ b/test/unit/conftest.py
@@ -1,4 +1,5 @@
 import sys
+import uuid
 sys.path.append(".")
 
 import pathlib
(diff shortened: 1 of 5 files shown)


WhiteMagic:
If you can't pick other input types and one should be able to, then yeah, that's definitely not intended. Though one can define what kind of inputs should be allowed, I'll have to look at how complete/buggy that is.

You may definitely run into bugs with the user scripts part as that's not been heavily tested and refactored quite heavily. So if you see something that's weird just let me know. The logical devices will definitely be part of the user scripts part as well but obviously right now aren't there yet. I should put that on the list of things to add.

test/unit/conftest.py
```diff
@@ -14,15 +15,19 @@
 import gremlin.event_handler
 
 
-def _make_fake_device(is_virtual: bool) -> dill.DeviceSummary:
-    """Creates a repeatable, faked DeviceSummary."""
-    # Data below was generated from a vJoy device.
-    guid = dill._GUID(
+def get_fake_device_guid(is_virtual: bool) -> uuid.UUID:
+    return dill._GUID(
```
WhiteMagic, line 19: Beware that while both `dill.GUID` and `uuid.UUID` are the same thing (as far as the values they can hold internally), their types aren't interchangeable, and it can cause tricky bugs if they're mixed. I try hard to have only `uuid.UUID` inside of Gremlin with `dill.GUID` propagating no further than the events Gremlin receives from the dill library. Though there are likely still places where `dill.GUID` are floating around.

code-monet: Good catch! The type hint was wrong, fixed.

code-monet:
> If you can't pick other input types and one should be able to

Ah, this was key. So it's user error in my part, I forgot that in my script I had specified it to be specifically an axis.

---

Okay, added coverage for the input variables for each type. Let me know if any more changes are needed.

WhiteMagic:
Looks good to me :+1:

PR #632, opened 2025-09-09
User script E2E integration test

code-monet:
Just device I/O coverage; the rest can/should be covered using the logical device in a separate test to be added later.

### DIFF
--- a/gremlin/input_cache.py
+++ b/gremlin/input_cache.py
@@ -91,10 +91,10 @@ class JoystickWrapper:
         def __init__(self, joystick_guid: uuid.UUID, index: int) -> None:
             super().__init__(joystick_guid, index)
 
-            self._value = types.HatDirection.Center
+            self._value = types.HatDirection.Center.value
 
         @property
-        def direction(self) -> types.HatDirection:
+        def direction(self) -> tuple[int, int]:
             return self._value
 
     def __init__(self, device_guid: uuid.UUID) -> None:
--- a/gremlin/user_script.py
+++ b/gremlin/user_script.py
@@ -485,9 +485,7 @@ class Script:
         }
 
         self._id = util.read_uuid(node, "script", "id")
-        self.path = _resolve_path(
-            Path(util.read_property(node, "path", PropertyType.String))
-        )
+        self.path = _resolve_path(util.read_property(node, "path", PropertyType.Path))
         self.name = util.read_property(node, "name", PropertyType.String)
 
         # Retrieve variable information from the script and instantiate them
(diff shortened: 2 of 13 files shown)


gremlin/input_cache.py
```diff
@@ -91,10 +91,10 @@ class Hat(Input):
         def __init__(self, joystick_guid: uuid.UUID, index: int) -> None:
             super().__init__(joystick_guid, index)
 
-            self._value = types.HatDirection.Center
```
WhiteMagic, line 94: The change is correct, though I think overall the original code is "wrong", i.e. there was a time when the `HatDirection` enum did not exist. But since it exists at this moment, I think it would be better to have the enum be the actual value rather than the tuple that is ambiguous in interpretation.

I think for this PR it makes sense to fix it up, but then we should go and exterminate `Tuple[int, int]` as the type being used for hat directions. Like with a few other things ideally the `EventListener` is where this "raw format" is last seen.

gremlin/user_script.py
```diff
@@ -478,7 +485,9 @@ def from_xml(self, node: ElementTree.Element) -> None:
         }
 
         self._id = util.read_uuid(node, "script", "id")
-        self.path = Path(util.read_property(node, "path", PropertyType.String))
+        self.path = _resolve_path(
```
WhiteMagic: It might be nicer to add a `PropertyType.Path` to do this in one go and not accidentally forget about it. It also has a bit of a self-documenting functionality. We don't have to enforce that the path is valid beyond what `Path()` does on it's own, i.e. no need for `exists` etc.

WhiteMagic:
The `PhysicalInputVariable` also has a `.decorator()` method which is the one I'd expect to be most widely used, the intent is that you can just write something like
```
mode_var = ModeVariable(...)
input_var = PhysicalInputVariable(...)

@input_var.decorator(mode_var) # Not fully sure right now if that was the right syntax but should be close to this
def some_callback(...):
    pass
```

gremlin/input_cache.py
code-monet: Sounds good, let me fix up that enum usage in my next PR.

gremlin/user_script.py
code-monet: Done, although I had to change `util.determine_value_type` to allow subclasses to be accepted - because while we would want to use `Path` in our code, the instances we actually get would be `WindowsPath`.

code-monet:
> The PhysicalInputVariable also has a .decorator() 

Ah that's much better than R13 - updated the example script.

gremlin/user_script.py
WhiteMagic: Great, and sub-classes being valid shouldn't cause any issues.

PR #634, opened 2025-09-12
Use the HatDirection enum everywhere in Gremlin

code-monet:
..except in the vjoy and dill wrappers

---

It's quite possible I didn't catch all the paths where it's used, but I at least verified:

1. Tests passing :)
2. Input viewer works
3. Macro action to and from hats work
4. Condition action works

### DIFF
--- a/action_plugins/condition/comparator.py
+++ b/action_plugins/condition/comparator.py
@@ -329,7 +329,7 @@ class DirectionComparator(AbstractComparator):
     def __call__(self, value: Value, events: List[event_handler.Event]) -> bool:
         # Check whether to use events or value
         if len(events) == 0:
-            return HatDirection.to_enum(value.current) in self.directions
+            return value.current in self.directions
 
         # Retrieve state of the events which should be just one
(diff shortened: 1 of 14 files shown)


WhiteMagic:
Great I'll look over that and see if I can remember any sneaky places the enum is used.

WhiteMagic:
Overall looks good to me. I think you found all the places it's being used at the moment.

Two things we could think of:
- `hat_tuple_to_direction` and `hat_direction_to_tuple` don't seem to be used anymore so should be safe to remove and declutter `utils`
- It might make sense to push HatDirection into vJoy as well, as the transformation into vJoy "values" happens in there based on tuple info. But that may also be a "later" thing as it could have quite a few ramifications.

code-monet:
Did both, verified by running the E2E integration tests and manually running a profile with macros (I should create an E2E for that also since it has a vJoy action...)

If it looks good please go ahead and merge.

WhiteMagic:
Great, thanks a lot!

PR #637, opened 2025-09-17
Add e2e integration test for relative axis

code-monet:


WhiteMagic:
Looks good to me, one warning about the risk of stepping too fast. DirectInput devices by their specification can report changes with up to 125Hz but not faster (I don't know if any exceed this but they are not meant to). So If we step manually too fast in an end to end test and expect changes to show up instantly we can run into possible issues as there are several "ticks" that are going on.

- Sending to vJoy
- vJoy being updated on the system level
- Dill picking the change up

So realistically we could be unlucky and see delays of probably up to 24ms or similar. I've never tried to find out what the actual delay is but I know that there are various rates at play and if we check really quickly we might "lag" behind.

code-monet:
Definitely a risk with integration testing to have flakiness because of random real-world delays etc. Hopefully between the `_assert_input_eventually_equals` function (which waits for up to 1 second by default for an input to attain a value), and the `patched_time` fixture introduced in this PR, we should not hit the race condition you described (but if the host is REALLY struggling for CPU we could exceed that 1 second timeout). If you ever see flakiness in any test, let me know, happy to take a look.

---

Interesting re: 125 Hz; I see some web hits on this topic but in my limited experience with (modern) racing wheels, they do send updates faster. My `FFFsake` plugin polls them at 5 Hz (could go faster but didn't feel the need to push it) and it gives meaningfully different values on each poll. The limit (125 Hz or whatever) is probably device dependent.

WhiteMagic:
For my T16000M it seems like 125Hz for the stick and something like 20Hz for the slider, according to this tool I found https://github.com/cakama3a/Polling

code-monet:
Interesting! For my Thrustmaster TMX wheel, the tool gives 429 Hz average with a minimum interval of less than 2 ms.

Issue #635, opened 2025-09-17
Race condition in map to vJoy relative axis thread


code-monet:
There's a race condition which can be reproduced basically 100% of the time by mapping a physical XY stick as follows:

X -> absolute vJoy X
Y -> relative vJoy Y

and then moving the physical stick diagonally. The thread will find that vJoy device 1 isn't acquired, tries to do so, but fails because the main thread is doing the same thing.

With these prints added to `vjoy.py`

<img width="1292" height="808" alt="Image" src="https://github.com/user-attachments/assets/7cd8c84b-07a5-48af-a07e-d06f9a50cd20" />

We see on the console:

```
acquire attempt, ID 1
acquire attempt, ID 1
acquire failed
Exception in thread Thread-9 (relative_axis_thread):
Traceback (most recent call last):
  File "C:\Program Files\Python312\Lib\threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "C:\Program Files\Python312\Lib\threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
acquired, ID 1
  File "c:\Users\Code Monet\Documents\GitHub\JoystickGremlin\action_plugins\map_to_vjoy\__init__.py", line 104, in relative_axis_thread
    vjoy_dev = VJoyProxy()[self.data.vjoy_device_id]
               ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\Users\Code Monet\Documents\GitHub\JoystickGremlin\vjoy\vjoy.py", line 953, in __getitem__
    raise e
  File "c:\Users\Code Monet\Documents\GitHub\JoystickGremlin\vjoy\vjoy.py", line 944, in __getitem__
    device = VJoy(index)
             ^^^^^^^^^^^
  File "c:\Users\Code Monet\Documents\GitHub\JoystickGremlin\vjoy\vjoy.py", line 532, in __init__
    raise VJoyError(
gremlin.error.VJoyError: 'Requested vJoy device is not available - vid: 1'
```

---

Proposal:

Let's acquire any vJoy devices with bindings in the loaded profile as soon as it is activated. This also fixes another issue I have - if the vJoy device cannot be acquired, you don't find out when the profile is activated, but rather when you press an input (and you might be inside a game at this point).

Alternatively we could use a lock in `VJoyProxy`.

WhiteMagic:
[...]

PR #638, opened 2025-09-17
Detect and handle concurrent vJoy instantiation attempt

code-monet:
Workaround for #635 

I couldn't replicate the bug anymore. The system log shows that the handler did get invoked:

```
2025-09-17 02:10:01      ERROR vJoy device 1 is already acquired by this process
NoneType: None
2025-09-17 02:10:01       INFO Concurrent instantiation attempt=1 for vJoy index=1, retrying...
```

(Re the `NoneType: None` it seems the `exception()` call in `error.GremlinError` is unable to find an exception to log...)

### DIFF
--- a/vjoy/vjoy.py
+++ b/vjoy/vjoy.py
@@ -27,7 +27,7 @@ import os
 
 from vjoy.vjoy_interface import VJoyState, VJoyInterface
 
-from gremlin.error import VJoyError
+from gremlin.error import VJoyConcurrencyError, VJoyError
 from gremlin.types import AxisNames, HatDirection
 import gremlin.spline
 
@@ -514,6 +514,11 @@ class VJoy:
 
         Args:
             vjoy_id: id of the vJoy device to initialize.
+
+        Raises:
+            VJoyConcurrencyError: If the requested vJoy device is already owned by this
+                process. Calling code should be able to catch and handle this.
+            VJoyError: If the requested vJoy device cannot be instantiated for any other reason.
         """
         self.vjoy_id = None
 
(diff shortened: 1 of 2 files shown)


WhiteMagic:
Do I understand the change correctly in that you try acquiring for a few times, hoping concurrent acquisitions resolve themselves, but give up if that doesn't happen? If so that seems sensible to me. One thing that came to mind, which the old code also fails to do is handle the case where the else/for clauses run their course without returning the device or failing. I suspect this should never be possible, but it might be good to just raise an error at the end that states this should never occur. At least we would know if it does happen for some crazy reason.

code-monet:
Good catch - I meant to add that `else` clause but forgot. Just added in the previous amended commit.

Yes, it is as you described. From what I can tell, the acquisition itself doesn't take as long as the rest of the initialization that happens after it. While it may still be possible for a race condition to happen during the device acquisition itself, it should be really unlikely. In the future, if needed, we could add a mutex around the acquisition part.

vjoy/vjoy.py
```diff
@@ -933,21 +936,33 @@ def __getitem__(self, index):
...
+                continue
             except VJoyError as e:
                 logging.getLogger("system").error(
                     f"Failed accessing vJoy id={index}, error is: {e}"
                 )
                 raise e
+            else:
+                VJoyProxy.vjoy_devices[index] = device
+                return device
+        else:
```
WhiteMagic: I think I wouldn't use this construct here. From my understanding we shouldn't be able to reach this so we should be able to just raise a vJoy or even Gremlin error saying that we reached this point which we shouldn't.

code-monet: Fair enough, since it's the end of the function - simplified.

PR #639, opened 2025-09-18
Add E2E integration test for macros - vjoy output actions only

code-monet:
Since `macro.py` has a bit of direct vJoy usage (and it didn't seem to make sense to try share code with `map_to_vjoy` action), I am adding a simple E2E integration test focused on the vJoy outputs part.

Rest of the macro should be tested using logical device I/O

### DIFF
--- a/test/integration/test_e2e_macro.py
+++ b/test/integration/test_e2e_macro.py
@@ -1,0 +1,125 @@
+# -*- coding: utf-8; -*-
+
+# Copyright (C) 2015 - 2025 Lionel Ott
+#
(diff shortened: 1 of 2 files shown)


WhiteMagic:
Looks good to me. A future addition might be the additional modes, though maybe that's better done with the LogicalDevice or the MacroManager rather than a full end-to-end test.

code-monet:
Agreed, I want to implement the rest of the test coverage with logical device so that those tests can be run on GitHub Actions.

PR #641, opened 2025-09-18
Fix bug where script manager UI was showing wrong scripts

code-monet:
The script manager sorts the scripts, but the model doesn't know that this happened.

AI suggested that the layout change signals are more efficient then doing a `modelReset` emit.


WhiteMagic:
In this instance it might be, as the layout signal mentions sorting as an example. In reality the correct solution is to figure out where the new script is being inserted and emit the beginInsertRows stuff for just that one "cell". But that's all kind of overkill for the amount of entries I'd expect to see.

PR #642, opened 2025-09-21
Add unit test coverage for user script variable XML load/save

code-monet:
Also fix bug where float variable would not accept integer literals for min/max/value


WhiteMagic:
:+1:

PR #648, opened 2025-10-09
Add auto mapper code and UI

### LINKED ISSUE #238: 1:1 Map only maps to vJoy device 1.
It should either ask what vJoy device to map to, or somehow know what vJoy device to map it to.

code-monet:
I leaned on AI for the UI and it's not great, but might be good enough to do the job. Open to suggestions now, or we can come back and improve it later.

Fixes #238

### DIFF
--- a/gremlin/auto_mapper.py
+++ b/gremlin/auto_mapper.py
@@ -1,0 +1,296 @@
+# -*- coding: utf-8; -*-
+
+# Copyright (C) 2015 - 2025 Lionel Ott
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Auto-mapping from physical DirectInput to vJoy devices.
+"""
+
+from collections.abc import Iterable
+import dataclasses
+import itertools
+import logging
+from typing import Self
+
+from action_plugins import map_to_vjoy, root
+import dill
+from gremlin import (
+    device_initialization,
+    mode_manager,
+    plugin_manager,
+    profile,
+    shared_state,
+    types,
+)
+from vjoy import vjoy
+
+
+@dataclasses.dataclass
+class AutoMapperOptions:
+    """Options for the auto-mapper.
+
+    More flags may be added here to further customize auto-mapper behavior.
+    """
+
+    repeat_vjoy_inputs: bool = False
+    overwrite_used_inputs: bool = False
+
+
+class AutoMapper:
+    """Generates "map to vJoy' actions for physical input devices.
+
+    The primary purpose is to help users with new profiles get started with
+    simple mappings for their input devices. The common use case is to map "available"
+    physical inputs to "available" vJoy inputs.
+
+    To keep things simple, a vJoy input is considered "available" even if there's a binding
+    to it in the profile, but:
+    1. The binding is from a disconnected device.
+    2. The binding is from a macro, a sub-action (usually temp, chain, or condition), or in a
+       user script.
+
+    A physical input is considered "available" if it has no binding in the profile. An option
+    is provided to overwrite unavailable inputs, in which case any existing bindings are removed,
+    unless they are direct vJoy mappings (which could have been created by this tool).
+
+    Bindings are generated for the current active mode only; bindings from other modes are
+    not checked (i.e. any physical and vJoy inputs used only in other modes are considered
+    available in the current mode).
+
+    This class should be instantiated after the current profile has been loaded/generated.
+    Functions should be called after device initialization is complete.
+    """
+
+    def __init__(self, profile: profile.Profile):
+        self._profile = profile
+        # For debug, testing and creating a report for the user.
+        self._created_mappings: list[map_to_vjoy.MapToVjoyData] = []
+
+    @classmethod
+    def from_current_profile(cls) -> Self:
+        return cls(shared_state.current_profile)
+
+    def generate_mappings(
+        self,
+        input_devices_guids: list[dill.GUID],
+        output_vjoy_ids: list[int],
+        options: AutoMapperOptions,
+    ) -> str:
+        """Generates mappings for the profile.
+
+        Args:
+            input_devices_guids: List of GUIDs representing the input devices to map from.
+            output_vjoy_ids: List of DeviceSummary objects representing the vJoy devices to map to.
+            options: Options for the auto-mapper.
+        
+        Returns:
+            A string report for the user summarizing new mappings.
+        """
+        input_devices = [
+            dev
+            for dev in device_initialization.physical_devices()
+            if dev.device_guid in input_devices_guids
+        ]
+
+        self._prepare_profile(input_devices, options)
+        used_vjoy_inputs = set(self._get_used_vjoy_inputs())
+        vjoy_axes = self._iter_unused_vjoy_axes(output_vjoy_ids, used_vjoy_inputs)
+        vjoy_buttons = self._iter_unused_vjoy_buttons(output_vjoy_ids, used_vjoy_inputs)
+        vjoy_hats = self._iter_unused_vjoy_hats(output_vjoy_ids, used_vjoy_inputs)
+        if options.repeat_vjoy_inputs:
+            vjoy_axes = itertools.cycle(vjoy_axes)
+            vjoy_buttons = itertools.cycle(vjoy_buttons)
+            vjoy_hats = itertools.cycle(vjoy_hats)
+        for physical_axis, vjoy_axis in zip(
+            self._iter_physical_axes(input_devices, options), vjoy_axes
+        ):
+            self._create_new_mapping(physical_axis, vjoy_axis)
+        for physical_button, vjoy_button in zip(
+            self._iter_physical_buttons(input_devices, options), vjoy_buttons
+        ):
+            self._create_new_mapping(physical_button, vjoy_button)
+        for physical_hat, vjoy_hat in zip(
+            self._iter_physical_hats(input_devices, options), vjoy_hats
+        ):
+            self._create_new_mapping(physical_hat, vjoy_hat)
+        return self._create_mappings_report()
+
+    def _prepare_profile(
+        self, input_devices: list[dill.DeviceSummary], options: AutoMapperOptions
+    ):
+        """Prepares the profile for an auto-map run."""
+        if options.overwrite_used_inputs:
+            for dev in input_devices:
+                self._profile.inputs.pop(dev.device_guid.uuid, None)
+
+    def _iter_physical_axes(
+        self, input_devices: list[dill.DeviceSummary], options: AutoMapperOptions
+    ) -> Iterable[profile.InputItem]:
+        """Iterates over physical axes that need to be mapped in a prepared profile."""
+        current_mode = mode_manager.ModeManager().current.name
+
+        for dev in input_devices:
+            for linear_index in range(dev.axis_count):
+                axis_index = dev.axis_map[linear_index].axis_index
+                input_item = self._profile.get_input_item(
+                    dev.device_guid.uuid,
+                    types.InputType.JoystickAxis,
+                    axis_index,
+                    current_mode,
+                    create_if_missing=True,
+                )
+                if not input_item.action_sequences:
+                    yield input_item
+
+    def _iter_physical_buttons(
+        self, input_devices: list[dill.DeviceSummary], options: AutoMapperOptions
+    ) -> Iterable[profile.InputItem]:
+        """Iterates over physical buttons that need to be mapped in a prepared profile."""
+        current_mode = mode_manager.ModeManager().current.name
+
+        for dev in input_devices:
+            for button in range(1, dev.button_count + 1):
+                input_item = self._profile.get_input_item(
+                    dev.device_guid.uuid,
+                    types.InputType.JoystickButton,
+                    button,
+                    current_mode,
+                    create_if_missing=True,
+                )
+                if not input_item.action_sequences:
+                    yield input_item
+
+    def _iter_physical_hats(
+        self, input_devices: list[dill.DeviceSummary], options: AutoMapperOptions
+    ) -> Iterable[profile.InputItem]:
+        """Iterates over physical hats that need to be mapped in a prepared profile."""
+        current_mode = mode_manager.ModeManager().current.name
+
+        for dev in input_devices:
+            for hat in range(1, dev.hat_count + 1):
+                input_item = self._profile.get_input_item(
+                    dev.device_guid.uuid,
+                    types.InputType.JoystickHat,
+                    hat,
+                    current_mode,
+                    create_if_missing=True,
+                )
+                if not input_item.action_sequences:
+                    yield input_item
+
+    def _get_used_vjoy_inputs(self) -> list[vjoy.VjoyInput]:
+        """Returns a list of all vJoy inputs that are already used in the prepared profile."""
+        used_vjoy_inputs = []
+        connected_device_uuids = set(
+            [dev.device_guid.uuid for dev in device_initialization.physical_devices()]
+        )
+        for device_uuid, input_items in self._profile.inputs.items():
+            if device_uuid not in connected_device_uuids:
+                logging.getLogger("system").debug(
+                    f"vJoy mappings from disconnected device {device_uuid=} "
+                    "are considered unused."
+                )
+                continue
+            for input_item in input_items:
+                if input_item.mode != mode_manager.ModeManager().current.name:
+                    logging.getLogger("system").debug(
+                        f"vJoy mapping from other mode {input_item.mode=} "
+                        "is considered unused."
+                    )
+                    continue
+                for binding in input_item.action_sequences:
+                    assert isinstance(binding.root_action, root.RootData)
+                    child_actions = binding.root_action.children
+                    for child_action in child_actions:
+                        if isinstance(child_action, map_to_vjoy.MapToVjoyData):
+                            used_vjoy_inputs.append(
+                                vjoy.VjoyInput(
+                                    child_action.vjoy_device_id,
+                                    child_action.vjoy_input_type,
+                                    child_action.vjoy_input_id,
+                                )
+                            )
+                        else:
+                            logging.getLogger("system").debug(
+                                f"Auto-mapper will reuse vJoy mappings in {child_action=}"
+                            )
+        return used_vjoy_inputs
+
+    def _iter_unused_vjoy_axes(
+        self, vjoy_ids: list[int], used_vjoy_inputs: set[vjoy.VjoyInput]
+    ) -> Iterable[vjoy.VjoyInput]:
+        """Returns a list of all vJoy inputs that are not used in the prepared profile."""
+        for vjoy_dev in device_initialization.vjoy_devices():
+            if vjoy_dev.vjoy_id not in vjoy_ids:
+                continue
+            for linear_index in range(vjoy_dev.axis_count):
+                axis_index = vjoy_dev.axis_map[linear_index].axis_index
+                vjoy_axis = vjoy.VjoyInput(
+                    vjoy_dev.vjoy_id, types.InputType.JoystickAxis, axis_index
+                )
+                if vjoy_axis not in used_vjoy_inputs:
+                    yield vjoy_axis
+
+    def _iter_unused_vjoy_buttons(
+        self, vjoy_ids: list[int], used_vjoy_inputs: set[vjoy.VjoyInput]
+    ) -> Iterable[vjoy.VjoyInput]:
+        for vjoy_dev in device_initialization.vjoy_devices():
+            if vjoy_dev.vjoy_id not in vjoy_ids:
+                continue
+            for button_id in range(1, vjoy_dev.button_count + 1):
+                vjoy_button = vjoy.VjoyInput(
+                    vjoy_dev.vjoy_id, types.InputType.JoystickButton, button_id
+                )
+                if vjoy_button not in used_vjoy_inputs:
+                    yield vjoy_button
+
+    def _iter_unused_vjoy_hats(
+        self, vjoy_ids: list[int], used_vjoy_inputs: set[vjoy.VjoyInput]
+    ) -> Iterable[vjoy.VjoyInput]:
+        for vjoy_dev in device_initialization.vjoy_devices():
+            if vjoy_dev.vjoy_id not in vjoy_ids:
+                continue
+            for hat_id in range(1, vjoy_dev.hat_count + 1):
+                vjoy_hat = vjoy.VjoyInput(
+                    vjoy_dev.vjoy_id, types.InputType.JoystickHat, hat_id
+                )
+                if vjoy_hat not in used_vjoy_inputs:
+                    yield vjoy_hat
+
+    def _create_new_mapping(
+        self, physical_input: profile.InputItem, vjoy_input: vjoy.VjoyInput
+    ):
+        """Creates a new mapping from physical_input to vjoy_input."""
+        p_manager = plugin_manager.PluginManager()
+        root_action = p_manager.create_instance(
+            root.RootData.name, physical_input.input_type
+        )
+        vjoy_action = p_manager.create_instance(
+            map_to_vjoy.MapToVjoyData.name, physical_input.input_type
+        )
+        vjoy_action.vjoy_device_id = vjoy_input.vjoy_id
+        vjoy_action.vjoy_input_id = vjoy_input.input_id
+        vjoy_action.vjoy_input_type = vjoy_input.input_type
+        root_action.insert_action(vjoy_action, "children")
+        binding = profile.InputItemBinding(physical_input)
+        binding.root_action = root_action
+        binding.behavior = physical_input.input_type
+        physical_input.action_sequences.append(binding)
+        self._created_mappings.append(vjoy_action)
+    
+    def _create_mappings_report(self) -> str:
+        """Creates a text report for the user after a mapping operation."""
+        return f"Created {len(self._created_mappings)} mappings."
--- a/gremlin/auto_mapper.py
+++ b/gremlin/auto_mapper.py
@@ -22,20 +22,11 @@ Auto-mapping from physical DirectInput to vJoy devices.
 from collections.abc import Iterable
 import dataclasses
 import itertools
-import logging
 from typing import Self
 
 from action_plugins import map_to_vjoy, root
 import dill
-from gremlin import (
-    device_initialization,
-    mode_manager,
-    plugin_manager,
-    profile,
-    shared_state,
-    types,
-)
-from vjoy import vjoy
+from gremlin import device_initialization, plugin_manager, profile, shared_state, types
 
 
 @dataclasses.dataclass
@@ -59,16 +50,16 @@ class AutoMapper:
     To keep things simple, a vJoy input is considered "available" even if there's a binding
     to it in the profile, but:
     1. The binding is from a disconnected device.
-    2. The binding is from a macro, a sub-action (usually temp, chain, or condition), or in a
-       user script.
+    2. The binding is not a direct vJoy mapping (e.g. from a macro, a sub-action like temp,
+       chain, or condition), or in a user script.
 
-    A physical input is considered "available" if it has no binding in the profile. An option
-    is provided to overwrite unavailable inputs, in which case any existing bindings are removed,
-    unless they are direct vJoy mappings (which could have been created by this tool).
+    A physical input is considered "available" if it has no binding/actions in the profile.
+    An option is provided to overwrite unavailable inputs, in which case any existing bindings
+    are removed.
 
-    Bindings are generated for the current active mode only; bindings from other modes are
+    Bindings are generated for the specified mode only; bindings from other modes are
     not checked (i.e. any physical and vJoy inputs used only in other modes are considered
-    available in the current mode).
+    available in the specified mode).
 
     This class should be instantiated after the current profile has been loaded/generated.
     Functions should be called after device initialization is complete.
@@ -85,6 +76,7 @@ class AutoMapper:
 
     def generate_mappings(
         self,
+        mode: str,
         input_devices_guids: list[dill.GUID],
         output_vjoy_ids: list[int],
         options: AutoMapperOptions,
@@ -92,10 +84,11 @@ class AutoMapper:
         """Generates mappings for the profile.
 
         Args:
+            mode: The mode to generate mappings for.
             input_devices_guids: List of GUIDs representing the input devices to map from.
             output_vjoy_ids: List of DeviceSummary objects representing the vJoy devices to map to.
             options: Options for the auto-mapper.
-        
+
         Returns:
             A string report for the user summarizing new mappings.
         """
@@ -106,7 +99,7 @@ class AutoMapper:
         ]
 
         self._prepare_profile(input_devices, options)
-        used_vjoy_inputs = set(self._get_used_vjoy_inputs())
+        used_vjoy_inputs = set(self._get_used_vjoy_inputs(mode))
         vjoy_axes = self._iter_unused_vjoy_axes(output_vjoy_ids, used_vjoy_inputs)
         vjoy_buttons = self._iter_unused_vjoy_buttons(output_vjoy_ids, used_vjoy_inputs)
         vjoy_hats = self._iter_unused_vjoy_hats(output_vjoy_ids, used_vjoy_inputs)
@@ -115,15 +108,15 @@ class AutoMapper:
             vjoy_buttons = itertools.cycle(vjoy_buttons)
             vjoy_hats = itertools.cycle(vjoy_hats)
         for physical_axis, vjoy_axis in zip(
-            self._iter_physical_axes(input_devices, options), vjoy_axes
+            self._iter_physical_axes(mode, input_devices, options), vjoy_axes
         ):
             self._create_new_mapping(physical_axis, vjoy_axis)
         for physical_button, vjoy_button in zip(
-            self._iter_physical_buttons(input_devices, options), vjoy_buttons
+            self._iter_physical_buttons(mode, input_devices, options), vjoy_buttons
         ):
             self._create_new_mapping(physical_button, vjoy_button)
         for physical_hat, vjoy_hat in zip(
-            self._iter_physical_hats(input_devices, options), vjoy_hats
+            self._iter_physical_hats(mode, input_devices, options), vjoy_hats
         ):
             self._create_new_mapping(physical_hat, vjoy_hat)
         return self._create_mappings_report()
@@ -137,11 +130,12 @@ class AutoMapper:
                 self._profile.inputs.pop(dev.device_guid.uuid, None)
 
     def _iter_physical_axes(
-        self, input_devices: list[dill.DeviceSummary], options: AutoMapperOptions
+        self,
+        mode: str,
+        input_devices: list[dill.DeviceSummary],
+        options: AutoMapperOptions,
     ) -> Iterable[profile.InputItem]:
         """Iterates over physical axes that need to be mapped in a prepared profile."""
-        current_mode = mode_manager.ModeManager().current.name
-
         for dev in input_devices:
             for linear_index in range(dev.axis_count):
                 axis_index = dev.axis_map[linear_index].axis_index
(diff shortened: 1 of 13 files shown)


gremlin/profile.py
```diff
@@ -722,12 +722,12 @@ def __init__(self, library: Library) -> None:
         Args:
             library: library instance that contains all action definitions
         """
-        self.device_id = None
-        self.input_type = None
-        self.input_id = None
-        self.mode = None
+        self.device_id: uuid.UUID | None = None
+        self.input_type: InputType | None = None
+        self.input_id: int | tuple[int, int] | None = None
```
WhiteMagic, line 726: Int is for joystick devices and the tuple for keyboard keys correct?

gremlin/ui/backend.py
```diff
@@ -312,6 +313,10 @@ def getInputItem(
     @Slot(result=LogicalDeviceManagementModel)
     def getLogicalDeviceManagementModel(self) -> LogicalDeviceManagementModel:
         return LogicalDeviceManagementModel(self)
+    
+    @Slot(result=AutoMapper)
```
WhiteMagic, line 334: Did you experiment with just creating the object directly in QML rather than going via the backend instance? That should likely work as usually the backend detour is only needed for objects that require some odd params during construction.

gremlin/ui/device.py
```diff
@@ -259,6 +260,7 @@ class DeviceListModel(QtCore.QAbstractListModel):
         "vid": lambda dev: "{:04X}".format(dev.vendor_id),
         "guid": lambda dev: str(dev.device_guid),
         "joy_id": lambda dev: dev.joystick_id,
+        "vjoy_id": lambda dev: dev.vjoy_id,
```
WhiteMagic, line 263: Is this safe for non vJoy devices?

qml/Main.qml
```diff
@@ -288,6 +292,23 @@ ApplicationWindow {
                 }
             }
 
+            ToolButton {
+                text: "1:1"
```
WhiteMagic, line 296: If this actually just renders the literal text "1:1" the whole bootstrap icon part isn't needed. That is only of use if you want to use an icon that's embedded in that font.

vjoy/vjoy.py
```diff
@@ -68,6 +68,14 @@ class HatType(enum.Enum):
     Continuous = 1
 
 
+class VjoyInput(NamedTuple):
```
WhiteMagic: As the vjoy module isn't using this on its own this isn't the most ideal place. Since it is only used with the auto_mapper feature I'd probably place it in that module. Technically we could reuse the `InputIdentifier` but that is tied to Qt so not actually useful for non-ui code.

If the thought is that there might be a need for a generic "deviceidentifier" class it would also be an option to put this into types and possible generalize to general joystick devices by using the `DeviceType` enum to encode the type of input (and extend it to include Logical as well as vjoy, joystick, and keyboard)

WhiteMagic:
I'll take a look at the mapper logic and UI stuff later. One thing I saw is that the UI here uses mostly row/column while I mostly tend to use rowlayout and collayout. I still don't like either as both kind of are dumb in different ways :-) For a simple UI both should be fine, I'll possibly even pull the whole thing into a separate branch to be able to actually run it as with UI code staring at the XML only tells you so much with QML.

WhiteMagic:
On the QML front, while it works, there are quite a few things I personally would change.

- Place options checkboxes below the device selections and replace them with Slider widgets (I'm not the biggest fan of them but that's where Windows is going)
- I try to use `_camelCase` for identifiers and only use them for components I need to refer to
- Use `RowLayout` and `ColumnLayout` instead of `Row` and `Column` as the anchor stuff gets quite messy and just contains magic values with no end
  - You can `anchors.fill: parent` in the main `ColumnLayout` and then have individual rows contain more `ColumnLayout` with `Layout.fillWidth: true` in both instances to get a 50/50 split.
-  The LLM also seemed to really like overly long identifier names
- Repeaters work but with a proper `AbstractListModel` it is better (usually) to use a `ListView` qml widget (there are instances where this doesn't work but I think this shouldn't be one of them)
- I don't have oversized buttons anywhere so there shouldn't be any here either
- Gremlin uses the Universal style (https://doc.qt.io/qt-6/qtquickcontrols-universal.html) and sticks to the colors it defines, one value that the page doesn't list is `Universa.baseLowColor` that is a useful "background" color. So I wouldn't define random hex values. For the `TextArea` I'd see if a simple line box is also suitable (such that it works on the same line as the non-oversized button) and then could also be turned into a separate "notification" widget so the style can be consistent and reused.
- I think having a short "description/manual" at the top explaining what/how to use the tool is fine. Alternatively you could add a `?` button on the bottom row with the button and status display.

gremlin/profile.py
code-monet: Correct!

vjoy/vjoy.py
code-monet: I was looking at `map_to_vjoy` and thought that this might be useful there. I wouldn't extend it to all devices since the device identifier type then gets overloaded, and same for input identifier (e.g. with keyboard key code tuples). Going by that, what's your preference between a) keeping it in vjoy and in the future re-use in `map_to_vjoy` (and `macro` perhaps), or b) move to `auto_mapper` (or c) move to `types` if you wish) ?

gremlin/ui/backend.py
code-monet: Changed.

I asked AI several times and it kept saying I had to go via `engine.rootContext().setContextProperty`... but now I see how you've been using `QtQml.QmlElement` and did the same thing.

gremlin/ui/device.py
code-monet: It's "safe", yes, you'll just get the placeholder -1 for those.

qml/Main.qml
code-monet: Remove that section.

code-monet:
Apologies that I intended to and forgot to attach a video of the UI - will do so for the next push so you don't have to spin up the UI each time. I'm fixing up the UI now; previously I did try to use the `Layout` variants of Row and Column, but the device checkboxes weren't clickable and didn't show any text, and I wasn't able to fix it. I'll try it again, referring to your other QMLs.

---

(Converting your list to checkboxes for my own tracking)

- [x]     Place options checkboxes below the device selections and replace them with Slider widgets (I'm not the biggest fan of them but that's where Windows is going)
- [x]     I try to use _camelCase for identifiers and only use them for components I need to refer to
- [x]     Use RowLayout and ColumnLayout instead of Row and Column as the anchor stuff gets quite messy and just contains magic values with no end
- [x]         You can anchors.fill: parent in the main ColumnLayout and then have individual rows contain more ColumnLayout with Layout.fillWidth: true in both instances to get a 50/50 split.
- [x]     The LLM also seemed to really like overly long identifier names
- [x]     Repeaters work but with a proper AbstractListModel it is better (usually) to use a ListView qml widget (there are instances where this doesn't work but I think this shouldn't be one of them)
- [x]     I don't have oversized buttons anywhere so there shouldn't be any here either
- [x]     Gremlin uses the Universal style (https://doc.qt.io/qt-6/qtquickcontrols-universal.html) and sticks to the colors it defines, one value that the page doesn't list is Universa.baseLowColor that is a useful "background" color. So I wouldn't define random hex values. For the TextArea I'd see if a simple line box is also suitable (such that it works on the same line as the non-oversized button) and then could also be turned into a separate "notification" widget so the style can be consistent and reused.
- [x]     I think having a short "description/manual" at the top explaining what/how to use the tool is fine. Alternatively you could add a ? button on the bottom row with the button and status display.

WhiteMagic:
No worries about the video it's easy enough to locally checkout the PR and run it.

For things not showing up when using layouts, there is a lot of weird quirky stuff going on with them at times. I often use the `DebugBox { target: _idOfSomething` } to visualize the space something takes up. With layouts it's important to give them size as normal with anchors but then their direct content needs to use `Layout.something` for dimensions as otherwise QML complains. The layouts also figure out how "big" something is based on the `implicitWidth` and `implicitHeight` of items. So sometimes it is necessary to compute those values and you can check if that's the issue by forcibly setting that value to something.

Also if you get stumped by it just commit what you have and let me know, I dislike UI stuff a lot but I've been dealing with a lot such that I know quite a few of the typical non-obvious ways to shoot yourself into the foot with them. Als LLMs seem to get better at knowing some of QML's quirks but sometimes they are also just totally wrong with their outputs.

code-monet:
Okay, I'm taking a short break but this is what I have so far (see checkboxes above). Additional notes:

1. I think you meant Switch and not Slider widget?
2. I didn't use baseLowColor for the notification box since it was matching the button, and the button color currently matches what I see e.g. in Actions

WhiteMagic:
Wow that's much nicer :-)

And yes I meant switch, I always get confused with what their name is in QML. Color use is fine, I just mentioned it in case it's useful because the documentation doesn't really say that the variable exists.

For the `ListView` you don't need to place them inside a `ScrollView` as that's doubling things up. `ListView`  is a scrollable container on its own. It behaves a bit dumb, so I'll actually make a `JGListView` or similar so I don't have to repeat the same things in them repeatedly.  The usual add-ons I do are:

```
ListView {
  clip: true // otherwise you can see the content outside the widget's boundariers

  // Always show a scrollbar, depends on use case
  ScrollBar.vertical: ScrollBar {
    policy: ScrollBar.AlwaysOn
  }
  // No bouncing or drag to move, i.e none of the phone behavior stuff
  flickableDirection: Flickable.VerticalFlick
  boundsBehavior: Flickable.StopAtBounds
}
```

For the `Switch` widgets you can also use the `checked` property of theirs in the `onToggled` which is directly the boolen you want to assign, i.e. no need to deal with the `position` and odd "truthness" things.

You don't need the `Rectangle` as the top element to get the spacing around the content. You can just add `anchors.margins: 10` to the to the top-level `ColumnLayout` to get the same effect.

The two `Switch` items could also be in a `RowLayout` of their own and split the space as they see fit, as opposed to being aligned with one of the two boxes, but that's really just another option and I don't have opinions on either layout.

Not all `Layout.fillX` are needed, but I honestly don't know the exact rules around what QML assumes as defaults or  the logic employed. There is this snippet in the documentation but it's kind of vague on what exactly will happen, especially when no `Layout.fillX` or `Layout.preferredX` is given.

> The [fillWidth](https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#fillWidth-attached-prop) and [fillHeight](https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#fillHeight-attached-prop) properties can either be true or false. If they are false, the item's size will be fixed to its preferred size. Otherwise, it will grow or shrink between its minimum and maximum size as the layout is resized. If there are multiple items with [fillWidth](https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#fillWidth-attached-prop) (or [fillHeight](https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#fillHeight-attached-prop)) set to true, the layout will grow or shrink the items relative to the ratio of their preferred size.

If you want I can also quickly implement the changes I mention into the QML file and probably just paste it here as making a PR against a PR might be funny, never tried that before.

**Edit:** Created that new `ListView` version `JGListView` which has the correct behavior by default and has a property `scrollbarAlwaysVisible` which allows toggling that on/off for the vertical scrollbar.

code-monet:
Done:

- [x] Using JGListView (and rebased on top of latest dev). Thanks for adding that!
- [x] Using `checked` property for `Switch`es
- [x] Removed top level Rectangle (AI apparently likes using more code than necessary)
- [ ] The Switches I left aligned with their respective device lists - that alignment/relationship makes sense to me
- [x] Removed some `FillWidth` and `FillHeight`s

---

I'm fine any way for QML/code suggestions - as a post in the PR, pasted snippet, separate branch that I can peek at - but really your posts have enough detail that I have been able to implement the changes without too much trouble :)

gremlin/profile.py
```diff
@@ -507,12 +507,12 @@ class Profile:
...
 
     def __init__(self) -> None:
-        self.inputs = {}
+        self.inputs: dict[uuid.UUID, list[InputItem]] = {}
         self.library = Library()
         self.settings = Settings(self)
         self.modes = ModeHierarchy(self)
         self.scripts = ScriptManager(self)
-        self.fpath = None
+        self.fpath: str | None = None
```
WhiteMagic, line 513: I'm really tempted to make a wrapper around Path only to be able to have an "invalid" path :-)

gremlin/ui/auto_mapper.py
```diff
@@ -0,0 +1,70 @@
...
+import logging
+
+from PySide6 import QtCore
+from PySide6 import QtQml
+
+import dill
+from gremlin import auto_mapper
+from gremlin.ui import backend
+
+QML_IMPORT_NAME = "Gremlin.UI"
```
WhiteMagic: I would suggest making a new package here, maybe `Gremlin.Tools` such that we have a nice group for everything that's tool-like. In the future the device remapper will need to live somewhere too and that'd be a good label I think.

gremlin/ui/auto_mapper.py
```diff
@@ -0,0 +1,70 @@
...
+QML_IMPORT_MAJOR_VERSION = 1
+
+
+@QtQml.QmlElement
+class AutoMapper(QtCore.QObject):
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+    @QtCore.Slot(dict, dict, bool, bool, result=str)
+    def create_mappings(self, physical_devices, vjoy_devices, overwrite, repeat) -> str:
```
WhiteMagic: Looking at this, since this is just a single function we really use, might also just put that into the `Backend` class. The Backend is getting bloated over time but that's kind of its purpose I guess.

If you needed more actual interaction with the UI and expose model data it would make sense to have a dedicated model, but since all you want to do is a "one shot" function execution without storing or exposing state I think the backend route is more "normal" for Gremlin. Though "normal" is changing over time too :-)

gremlin/ui/auto_mapper.py
```diff
@@ -0,0 +1,70 @@
...
+        Args:
+            physical_devices: Dictionary of {device_guid: is_selected} for physical devices
+            vjoy_devices: Dictionary of {vjoy_id: is_selected} for vJoy devices
+            overwrite: Whether to overwrite existing mappings
+            repeat: Whether to repeat vJoy mappings
+        
+        Returns:
+            A string report for the user summarizing new mappings.
+        """
+        logging.getLogger("system").info(
```
WhiteMagic, line 52: I tend to use f"{variable}" strings for formatting everywhere as opposed to the really old "%s" substitution style.

gremlin/ui/backend.py
WhiteMagic: The `QtQML.QmlElement` decorator is a shortcut to exporting a QML object to the engine. It does what the `PluginManager` does
```
QtQml.qmlRegisterType(plugin.create.model, "Gremlin.ActionPlugins", 1, 0, plugin.create.model.__name__)
```
with the decorator. Though I may at some point replace all of them with the specific format as linters don't understand what goes on with that decorator and then give you all kinds of false positive complaints.

gremlin/ui/device.py
WhiteMagic: Ok as long as it doesn't horribly crash, though I also don't think that value should be queried by anything that doesn't already know it deals with vJoy devices.

qml/TextOutputBox.qml
```diff
@@ -0,0 +1,36 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import QtQuick
+import QtQuick.Controls
+import QtQuick.Controls.Universal
+
+Rectangle {
+    id: root
+    property alias text: _notificationLabel.text
+
```
WhiteMagic, line 25: Possibly good to add 
```
implicitHeight: _notificationLabel.height
implicitWidth: _notificationLabel.width
```
What this does is make the Rectangle have the size large enough to show the content of the label. Without that it won't have a size, i.e. nothing shown, unless the user specifies one. Using the `implicit` values allows Layouts to use that information as a hint for the required size when they do their layouting.

So what should happen in your code if you remove the `Layout.fillWidth` etc. is that the text should still show up without overlapping other things because it knows how big it is. Without the additional information you actually wouldn't see the text. At least that's what I think should happen, which with QML oftentimes isn't super clear.

vjoy/vjoy.py
WhiteMagic: Keeping it to vJoy is fine since it's a bit of an odd thing. I would not put it in `auto_mapper` as that's not something that shouldn't need to be imported just to get this type. I think `types` might be the cleanest place for this. My main reason for not wanting it inside of `vjoy` is that I'd like to keep that module as Gremlin free as possible. Such that if in the future it is replaced or another direct input emulation driver comes along there aren't many threads to untangle.

Since this "type" is just a named tuple it won't introduce crazy import dependencies which could cause circular imports, which also makes it a safe candidate for types.

WhiteMagic:
> Done:
> 
>     * [x]  Using JGListView (and rebased on top of latest dev). Thanks for adding that!
> 
>     * [x]  Using `checked` property for `Switch`es
> 
>     * [x]  Removed top level Rectangle (AI apparently likes using more code than necessary)
> 
>     * [ ]  The Switches I left aligned with their respective device lists - that alignment/relationship makes sense to me
> 
>     * [x]  Removed some `FillWidth` and `FillHeight`s
> 
> 
> I'm fine any way for QML/code suggestions - as a post in the PR, pasted snippet, separate branch that I can peek at - but really your posts have enough detail that I have been able to implement the changes without too much trouble :)

Nice that looks good to me. I'll try to go over the main auto mapper code in the evening, didn't manage to look at that one yet.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+        """Prepares the profile for an auto-map run."""
+        if options.overwrite_used_inputs:
+            for dev in input_devices:
+                self._profile.inputs.pop(dev.device_guid.uuid, None)
+
+    def _iter_physical_axes(
+        self, input_devices: list[dill.DeviceSummary], options: AutoMapperOptions
+    ) -> Iterable[profile.InputItem]:
+        """Iterates over physical axes that need to be mapped in a prepared profile."""
+        current_mode = mode_manager.ModeManager().current.name
```
WhiteMagic: This won't give you the currently selected mode in the UI. The mode manager only serves a function when Gremlin is running. The UI selection information is only held in the `UIState` object but that one exists only as an instance within the UI and isn't exposed in any way.

In short, there is no way at the moment for code to know what the currently selected mode is without the UI passing that information on. If that information is needed I see two ways:
1. Devise a mechanism to expose UI state to Gremlin code as well.
2. Have the UI pass the name of the active mode on to code it calls.

Realistically most code shouldn't need to be able to query the UI state, though here I could see that being necessary. Though exposing it could become quite ugly so I probably lean more towards letting the UI forward that information.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+                    current_mode,
+                    create_if_missing=True,
+                )
+                if not input_item.action_sequences:
+                    yield input_item
+
+    def _get_used_vjoy_inputs(self) -> list[vjoy.VjoyInput]:
+        """Returns a list of all vJoy inputs that are already used in the prepared profile."""
+        used_vjoy_inputs = []
+        connected_device_uuids = set(
```
WhiteMagic: Did you run across instances where a device is listed multiple times in that list? If so that's a bug as it makes no sense for the same device to exist more than once. So `set` should be superfluous, if it's needed the other code has bugs.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+            [dev.device_guid.uuid for dev in device_initialization.physical_devices()]
+        )
+        for device_uuid, input_items in self._profile.inputs.items():
+            if device_uuid not in connected_device_uuids:
+                logging.getLogger("system").debug(
```
WhiteMagic: This may become very spammy and I'd probably just skip this.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+                    f"vJoy mappings from disconnected device {device_uuid=} "
+                    "are considered unused."
+                )
+                continue
+            for input_item in input_items:
+                if input_item.mode != mode_manager.ModeManager().current.name:
```
WhiteMagic: As mentioned above, this is not the same as the mode selected in the UI. Another option would be to have a mode selector dropdown in the 1:1 mapping dialog. That way it's actually clear to which mode the mappings are being added, as opposed to the user needing to figure out that it's the currently selected one.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+                    logging.getLogger("system").debug(
+                        f"vJoy mapping from other mode {input_item.mode=} "
+                        "is considered unused."
+                    )
+                    continue
+                for binding in input_item.action_sequences:
+                    assert isinstance(binding.root_action, root.RootData)
```
WhiteMagic, line 200: Unless someone manually messed with the XML you can't actually have the root_action be anything other than Root. Though there is nothing saying there couldn't be other options in the future if there is a need.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+                    child_actions = binding.root_action.children
+                    for child_action in child_actions:
```
WhiteMagic: Could be simplified to: `for child_action in binding.root_action.children:`.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+                                vjoy.VjoyInput(
+                                    child_action.vjoy_device_id,
+                                    child_action.vjoy_input_type,
+                                    child_action.vjoy_input_id,
+                                )
+                            )
+                        else:
+                            logging.getLogger("system").debug(
+                                f"Auto-mapper will reuse vJoy mappings in {child_action=}"
+                            )
```
WhiteMagic: Not sure what the log is meant to mean, and similar to the earlier one would just result in a lot of spam in the log file.

I think there is also a logic error here. The code assumes that all vJoy maps are children of the root node, that is not true. In R14 you can have arbitrarily deep nestings and each action may have "children", e.g. the condition action has children actions it runs when the condition holds and another set when they don't. Using `children` on the root action is also using knowledge/information that may change.

The `base_classes.AbstractActionData` class, from which every action has to inherit, does provide the basic building block you'd have to use to get all child actions of a given action via the `get_actions(selector: str|None=None)`. If no argument is given it will return all children of that action, irrespective of the "child container" they live in. That interface must be implemented by every action that can have children, otherwise the UI breaks, so you using that will always give you the right actions. Now what this does not do is recursively go through the children you get in that way and get their children, if they  have any. So if you have the root action the following should give you all actions that are in the possibly very deep nesting of actions:

```
all_actions = []
candidate_actions = root_action.get_actions()
while len(candidate_actions) > 0:
    all_actions.append(candidate_actions.pop(0))
    candidate_actions.extend(all_action[-1].get_children())
```
This should be correct but testing wouldn't hurt :-)

Once you have the list of all actions you can filter them or you could already filter them in there if you wanted. For the filtering there is another approach you could use. The `Library` has a method `actions_by_type` which returns all action data instances that match a given type. You could then use that to do a `action in vjoy_actions` check, though tbh. I have no idea which one is faster or more convenient.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+        return used_vjoy_inputs
+
+    def _iter_unused_vjoy_axes(
+        self, vjoy_ids: list[int], used_vjoy_inputs: set[vjoy.VjoyInput]
+    ) -> Iterable[vjoy.VjoyInput]:
+        """Returns a list of all vJoy inputs that are not used in the prepared profile."""
+        for vjoy_dev in device_initialization.vjoy_devices():
+            if vjoy_dev.vjoy_id not in vjoy_ids:
```
WhiteMagic, line 219: Do you know what happens with this if someone has vjoy device 1 and 3 but not 2? It's an odd edge case but some people do this for reasons.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+                continue
+            for linear_index in range(vjoy_dev.axis_count):
+                axis_index = vjoy_dev.axis_map[linear_index].axis_index
+                vjoy_axis = vjoy.VjoyInput(
+                    vjoy_dev.vjoy_id, types.InputType.JoystickAxis, axis_index
+                )
+                if vjoy_axis not in used_vjoy_inputs:
+                    yield vjoy_axis
```
WhiteMagic, line 227: How does this behave when you can't yield because you "ran out of" unused inputs?

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+        )
+        vjoy_action.vjoy_device_id = vjoy_input.vjoy_id
+        vjoy_action.vjoy_input_id = vjoy_input.input_id
+        vjoy_action.vjoy_input_type = vjoy_input.input_type
+        root_action.insert_action(vjoy_action, "children")
+        binding = profile.InputItemBinding(physical_input)
+        binding.root_action = root_action
+        binding.behavior = physical_input.input_type
+        physical_input.action_sequences.append(binding)
+        self._created_mappings.append(vjoy_action)
```
WhiteMagic: As this is mostly a copy of what `gremlin.ui.profile.InputItemModel.newActionSequence` does I think it might make sense to move the logic of creating a new `InputBinding` into the actual `gremlin.profile.InputItem` class. That way we remove the duplication and more importantly if there is ever anything wrong with it or changed both uses are automatically fixed.

Moving this should be straight forward I think. In the `InputItemModel` code the call to the new method just needs to be in between the `beginInsertRows()` and `endInsertRows()` calls.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+from action_plugins import map_to_vjoy, root
+import dill
+from gremlin import (
+    device_initialization,
+    mode_manager,
+    plugin_manager,
+    profile,
+    shared_state,
+    types,
+)
```
WhiteMagic: I've mostly just done list without the tuple notation and broken it with `\` when it exceeded 80 chars in length.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+
+@dataclasses.dataclass
+class AutoMapperOptions:
+    """Options for the auto-mapper.
+
+    More flags may be added here to further customize auto-mapper behavior.
+    """
+
+    repeat_vjoy_inputs: bool = False
+    overwrite_used_inputs: bool = False
```
WhiteMagic, line 40: To me it feels a bit overkill to use a class for just two params. However, if we'd need to pass the mode name around as well then I think it starts to be sensible to pack them into one single thing.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+
+    The primary purpose is to help users with new profiles get started with
+    simple mappings for their input devices. The common use case is to map "available"
+    physical inputs to "available" vJoy inputs.
+
+    To keep things simple, a vJoy input is considered "available" even if there's a binding
+    to it in the profile, but:
+    1. The binding is from a disconnected device.
+    2. The binding is from a macro, a sub-action (usually temp, chain, or condition), or in a
+       user script.
```
WhiteMagic: Hmm this case I probably wouldn't exclude. If someone has those they may not accidentally want them to be used. But only if its part of an active device obviously.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+
+    A physical input is considered "available" if it has no binding in the profile. An option
+    is provided to overwrite unavailable inputs, in which case any existing bindings are removed,
+    unless they are direct vJoy mappings (which could have been created by this tool).
```
WhiteMagic: Just a thought occurred, you need to check that the root action has 0 children, if someone clicked on an item sometimes Gremlin initializes things but they aren't "there", so assuming anything that has an `InputItemBinding` is used may not be true.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+
+    Bindings are generated for the current active mode only; bindings from other modes are
+    not checked (i.e. any physical and vJoy inputs used only in other modes are considered
+    available in the current mode).
```
WhiteMagic: As mentioned below, the code can't do this at the moment. I ran into the same problem when working on the condition action again. Only there I actually don't care and could ignore this issue.

gremlin/ui/auto_mapper.py
WhiteMagic: This should be `createMappings` the reason is that Qt uses `camelCase` for everything and while it's not the usual Python naming convention I use it when PySide6 code exposes something as a `Slot` or `Property`. That way it's directly clear that the method deals with QML code and on the QML side it doesn't look out of place.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,296 @@
...
+            options: Options for the auto-mapper.
+        
+        Returns:
+            A string report for the user summarizing new mappings.
+        """
+        input_devices = [
+            dev
+            for dev in device_initialization.physical_devices()
+            if dev.device_guid in input_devices_guids
+        ]
```
WhiteMagic, line 97: Ah the `DeviceSummary` objects are using `dill.GUID` instead of `uuid.UUID` correct? I wouldn't change that only asking so no awkward bugs get in.

gremlin/ui/auto_mapper.py
```diff
@@ -0,0 +1,70 @@
...
+        return mapper.generate_mappings(
+            [
+                dill.GUID.from_str(guid)
+                for (guid, chosen) in physical_devices.items()
+                if chosen
+            ],
+            [int(vjoy_id) for (vjoy_id, chosen) in vjoy_devices.items() if chosen],
+            auto_mapper.AutoMapperOptions(
+                repeat_vjoy_inputs=repeat,
+                overwrite_used_inputs=overwrite,
```
WhiteMagic: Do you have to use named parameters or does it also allow you to just use them in the positional order?

WhiteMagic:
Ok had a look over the mapping code, overall looks correct and good to me, despite the lengthy comments :-)

gremlin/ui/backend.py
code-monet: Ack (no action needed, right?)

vjoy/vjoy.py
code-monet: Sounds reasonable, moved to `types`.

gremlin/ui/auto_mapper.py
code-monet: Good idea, done.

gremlin/ui/auto_mapper.py
code-monet: That would be less code, but I agree with your concern for bloat in the Backend, which is not just a "backend" but also serving the purpose of a namespace. I'd keep this as a separate file just to keep things cleaner and reduce how many things are imported into `backend.py`.

And who knows, we might end up making the auto_mapper implementation/UI more complicated/feature-full in the future!

gremlin/ui/auto_mapper.py
code-monet: There is a performance benefit to doing it the old way in logging calls - the string isn't formatted unless the log line actually needs to be generated.

However since Gremlin doesn't really change log levels this aspect won't come into play - so I have changed it to f-strings.

qml/TextOutputBox.qml
code-monet: I have added these, but it didn't have the desired effect: `layout.fillWidth` is still needed. Per AI, this is expected:

"The issue is that the TextOutputBox relies on its implicit dimensions, but in a RowLayout, the layout needs explicit constraints to properly size the component. 

Without an explicit Layout.preferredHeight, the RowLayout collapses the height to 0 because:
* The TextOutputBox's implicit height depends on its content
* But the content's height depends on the component's height (circular dependency)
* The layout can't resolve this, so it collapses to 0 height"

gremlin/auto_mapper.py
code-monet: Not needed, this is a leftover from earlier logic - removed set cast.

gremlin/auto_mapper.py
code-monet: Removed.

gremlin/auto_mapper.py
code-monet: I just like having such a lightweight internal check wherever there's an assumption being made about a type (that the type checker can't determine). It makes error messages easier to understand (mainly intended for developers).

gremlin/auto_mapper.py
code-monet: Done.

gremlin/auto_mapper.py
code-monet: Replaced log with comment.

I realized my class document didn't match implementation, updated that; based off our earlier conversation in #238 I decided to keep things simple and only consider a vJoy input "used" if it was in a direct mapping. The idea being, a user could run the tool, remove/change some of the generated bindings, and re-run the tool.

For more complicated logic, I propose punting them to the future until someone asks for it (and probably adding a switch in the UI, like we have for overwriting existing physical inputs).

gremlin/auto_mapper.py
code-monet: It should be fine, at least the UI is correctly showing and passing the VJoy device ID in such a configuration, and none of the code here makes an assumption about continuity (at least going by `range` calls...)

gremlin/auto_mapper.py
code-monet: No more bindings will be created for that input type, unless "repeat VJoy devices" is checked, in which case the inputs will be looped as discussed in #238 

Tangentially, I realize that a lot of edge cases are possible and that the unit tests don't really cover many of those... I can come back and add more unit tests in a future PR, using mock devices rather than a profile (which for the time being is nice, because I can load it up and view it in Gremlin).

gremlin/auto_mapper.py
code-monet: Changed, although the `black` formatter prefers this. What formatter are you using? We could specify it in `contributingmd` and configure it via the TOML, e.g. https://developer.lsst.io/python/formatting.html#id1

gremlin/auto_mapper.py
code-monet: Indeed I did it with a class because I expect we'll end up adding more things to it.

gremlin/auto_mapper.py
code-monet: I had it coded up like that originally, but simplified it because it didn't seem like a thing that could happen. Perhaps instead of accommodating that here, would it be better to track down situations where empty InputItemBindings are created and prevent that from happening?

gremlin/auto_mapper.py
code-monet: I'll put mode into a drop down in the tool UI.

gremlin/ui/auto_mapper.py
code-monet: Renamed.

gremlin/auto_mapper.py
code-monet: Yes, which is probably fine - but in a future PR I can switch `device_initialization` to use UUID instead.

gremlin/ui/auto_mapper.py
code-monet: Positional will also work, switched to that.

gremlin/profile.py
code-monet: In theory the `exists()` function is what you'd use to determine if it's "valid", whereas there's still a separate concept, of whether one has been specified (yet), for which using `| None` is fine, IMO.

gremlin/auto_mapper.py
code-monet: I will follow your suggestion of adding mode to a drop down in the UI.

gremlin/auto_mapper.py
code-monet: Good call, added `add_item_binding`

gremlin/auto_mapper.py
code-monet: I had coded that up partially earlier, but then removed it because it was getting complicated (also needed knowledge of how macro, chain etc using vJoy). Per our [earlier thread](https://github.com/WhiteMagic/JoystickGremlin/issues/238#issuecomment-3337092915) perhaps we can come back and modify this later if needed? I agree with our earlier discussion that it's unlikely someone will add somewhat advanced bindings and *then* run the tool.

gremlin/auto_mapper.py
code-monet: Good idea, added drop down in the tool UI.

code-monet:
Okay, I *think* I have responded to all comments - let me know what you think, I can make more changes depending on what we decided we want to put into this first implementation.

I re-used the TextOutputBox but something is still missing because it needs either `fillWidth` or `preferredWidth`, otherwise it collapses to 0.

gremlin/auto_mapper.py
WhiteMagic: Yes for now this is totally fine. In the future it's probably easiest/safest to not even expose the GUID via the `DeviceSummary`. That way it can't "escape" the `dill` module.

gremlin/auto_mapper.py
WhiteMagic: I generally agree, though here it is currently an invariant that must hold, but in the future this may not be true. I guess it might be fine to run into this issue at that time, though.

gremlin/auto_mapper.py
WhiteMagic: Ok then that should be fine :+1:

gremlin/auto_mapper.py
WhiteMagic: Sounds good to me, no need to overdo the bullet proofing :-)

gremlin/profile.py
WhiteMagic: The trouble is with an empty path becoming just `.` rather than being invalid as is the case in C++ :-( But yeah it's just a general annoyance but not too common to fix ... for now at least.

gremlin/ui/auto_mapper.py
WhiteMagic: I don't tend to care for performance aspects unless it's a very obvious case where it will matter, so usually tight loops of compute or reacting to user inputs in even handling. This being a tool, I think there can be a lot of "slow" code and no one would notice.

gremlin/ui/backend.py
WhiteMagic: Yep all good, I just added the explanation in case you were wondering what that thing does.

qml/TextOutputBox.qml
WhiteMagic: I'd say the LLM is inventing things there, the documentation about layouts has this to say.

> Note: Do not bind to the x, y, width, or height properties of items in a layout, as this would conflict with the goals of Layout, and can also cause binding loops. The width and height properties are used by the layout engine to store the current size of items as calculated from the minimum/preferred/maximum attached properties, and can be ovewritten each time the items are laid out. Use [Layout.preferredWidth](https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#preferredWidth-attached-prop) and [Layout.preferredHeight](https://doc.qt.io/qt-6/qml-qtquick-layouts-layout.html#preferredHeight-attached-prop), or [implicitWidth](https://doc.qt.io/qt-6/qml-qtquick-item.html#implicitWidth-prop) and [implicitHeight](https://doc.qt.io/qt-6/qml-qtquick-item.html#implicitHeight-prop) to specify the preferred size of items.

However, once you start nesting things the lines get very blurry and stuff might just misbehave and I have mostly given up trying to fully understand the rhyme and reason behind the behavior. If it doesn't work we can omit it or leave it in in the hopes that maybe it works in other situations. Though realistically in most cases one will specify dimensions anyhow.

So no need to waste time trying to make it behave.

WhiteMagic:
I had one other though on the ui mode bit. It could be possible to store the UI mode also in the `ModeManager` as it is Qt enabled. However, the drawback/ugliness of this is that Qt code gets dragged into otherwise Qt free code. I try to keep everything outside the `gremlin/ui` folder free of Qt stuff.

gremlin/auto_mapper.py
```diff
@@ -0,0 +1,276 @@
...
+import itertools
+from typing import Self
+
+from action_plugins import map_to_vjoy, root
+import dill
+from gremlin import device_initialization, plugin_manager, profile, shared_state, types
+
+
+@dataclasses.dataclass
+class AutoMapperOptions:
```
WhiteMagic, line 33: Since we have this data class I'd put the `mode` value into this as well, otherwise, it's a bit odd to pass a param struct but then still need to pass around more params outside of it.

gremlin/profile.py
code-monet: Definitely re: empty strings. Marking as resolved, something to consider in the future.

qml/TextOutputBox.qml
code-monet: I'll leave it in, perhaps one day in a different QML we'll find that it works and then we'll have solved the mystery!

gremlin/auto_mapper.py
code-monet: Thanks, marking as resolved.

gremlin/auto_mapper.py
code-monet: If we implement the ability to copy bindings from one device to another, then I can't think of any other good reason to make the GUID available to the user. (marking as resolved).

gremlin/auto_mapper.py
code-monet: Moved to options.

code-monet:
>  It could be possible to store the UI mode also in the ModeManager as it is Qt enabled. However, the drawback/ugliness of this is that Qt code gets dragged into otherwise Qt free code. I try to keep everything outside the gremlin/ui folder free of Qt stuff.

I don't use the ModeManager anymore, the dropdown idea is good! More generally, it seems reasonable to keep UI and non-UI stuff separated.

WhiteMagic:
Very nice, good to merge from my side. Let me know if you wanted to add anything else in, otherwise I'll merge it.

code-monet:
Go ahead with the merge, I'll send a new PR with some minor improvements.

PR #651, opened 2025-10-26
Add more detail to the automapper result messaging

code-monet:
Examples:

image

image


WhiteMagic:
lgtm

code-monet:
I think the number of retaining bindings might be inaccurate if we run out of vjoy inputs. It might be useful to add a sort of summary popup one day.

WhiteMagic:
Could be an idea, though I'd keep it in the summary dialog or similar, I tend to dislike popups for informative things.

Issue #447, opened 2022-08-20
Improve swap devices utility


WhiteMagic:
Reimplementing the swap devices utility to QML is needed with some bug fixes and a renaming. While the utility was intended to take a configuration and copy it to other devices, the primary use is when people move to a new machine or similar. As such, renaming the tool might make sense. There is also a bug resulting in the description of actions not being copied. This likely won't be an issue as the implementation will have to reimplement large parts of the copy logic.

code-monet:
Looking for other ways to help out - how about this one? There are some parallels with the automapper code and GUI which would help. I was thinking two main cases should be supported:

1. New machine but same device - basically a change of GUID only.
2. Different device - copy over axes, buttons and hats until either device runs out of inputs, or no more existing bindings.

WhiteMagic:
Those are the two ways in which it is employed, yes. For the most part this is quite easy to accomplish as the inputs section of the XML containing `profile.InputItem` data is what makes the bulk of the changes. Though because nothing is ever easy there are a few other places where device information crops up. At a minimum these places also get device data (GUID really)
- condition
- macro
- merge axis
- user scripts

The somewhat good thing is that in R14 all actions should use the `PropertyType.UUID` thing in the XML serialization to make UUID data marked. Though I'm not keen on having to mangle XML data in code for this feature, so I would ignore the fact this exists.

There are two ways I can see how to make this feature work:
1. The swap functionality just has hard-coded information about what actions and fields within to transform.
2. The abstract base class of actions is modified to expose functionality for swapping them over and register/inform that they need to be processed.

Option 1 is easier in the short them but option 2 is more general and scales nicer but also requires dealing with a lot of the actions and making things discoverable etc. Also in any case there will always be special cases like the user scripts that probably always will have to be manual code. Though for an initial version they might also be ignored.

As for the conversion, I probably would disallow mapping from device A to device B if mappings on device A require inputs that device B does not have, i.e. you require Button 73 but that doesn't exist, you can't map, same for axes. This is just to avoid awkward partial conversions etc. These partial swaps are what improvements to the library UI should cover. The library should be (at some point) "browsable" like a set of actions from which you can drag items onto inputs to pick and choose what you want.

So the device swap should only support actual swaps but no partials as that makes it complicated, harder to understand, and there will be a different system for partial copies.

WhiteMagic:
[...]

PR #654, opened 2025-11-15
Add device swap tool

code-monet:
Adds core library and UI

Per our discussion in #447  this is currently quite simple. We don't check for cases where the swap will create invalid bindings, like:

> Obviously there is the issue of not knowing how many inputs exist on a disconnected device

Let me know if you'd like to see more features or checks.

UI with source device disconnected:
SwapDevicesDisconnected

and when connected:

SwapDevicesConnected

I'll add more unit test coverage in the next few days.

### DIFF
--- a/action_plugins/condition/condition.py
+++ b/action_plugins/condition/condition.py
@@ -491,19 +491,20 @@ class JoystickCondition(AbstractCondition):
             input_type: InputType,
             input_id: int
         ) -> None:
-            self.device_uuid = device_uuid
+            self.device_uuid: uuid.UUID
             self.input_type = input_type
             self.input_id = input_id
             self.joystick = None
-            self.update_for_device()
+            self.initialize_for_uuid(device_uuid)
         
-        def update_for_device(self):
+        def initialize_for_uuid(self, device_uuid: uuid.UUID):
             try:
-                self.joystick = Joystick()[self.device_uuid]
+                self.joystick = Joystick()[device_uuid]
             except error.GremlinError:
                 pass
+            self.device_uuid = device_uuid
             self.device_lookup = DeviceDatabase().get_mapping_by_uuid(
-                self.device_uuid
+                device_uuid
             )
 
         def get(self) -> bool | float | HatDirection:
@@ -619,8 +620,7 @@ class JoystickCondition(AbstractCondition):
         performed_swap = False
         for state in self._states:
             if state.device_uuid == old_uuid:
-                state.device_uuid = new_uuid
-                state.update_for_device()
+                state.initialize_for_uuid(new_uuid)
                 performed_swap = True
         return performed_swap
 
--- a/action_plugins/condition/condition.py
+++ b/action_plugins/condition/condition.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABCMeta, abstractmethod
-from typing import Any, List, Optional, TYPE_CHECKING
+from typing import Any, List, Optional, override, TYPE_CHECKING
 import uuid
 from xml.etree import ElementTree
 
@@ -255,6 +255,10 @@ class AbstractCondition(QtCore.QObject):
             self._states = state_list
             self.statesChanged.emit(self.states)
 
+    def swap_uuid(self, old_uuid: uuid.UUID, new_uuid: uuid.UUID) -> bool:
+        """Swaps occurrences of the old UUID with the new one for this condition."""
+        return False
+
 
 @QtQml.QmlElement
 class VJoyCondition(AbstractCondition):
@@ -491,6 +495,9 @@ class JoystickCondition(AbstractCondition):
             self.input_type = input_type
             self.input_id = input_id
             self.joystick = None
+            self.update_for_device()
+        
+        def update_for_device(self):
             try:
                 self.joystick = Joystick()[self.device_uuid]
             except error.GremlinError:
@@ -606,6 +613,16 @@ class JoystickCondition(AbstractCondition):
         # No need to change the comparator as we don't rely on the input's
         # type for condition checks.
         pass
+
+    @override
+    def swap_uuid(self, old_uuid: uuid.UUID, new_uuid: uuid.UUID) -> bool:
+        performed_swap = False
+        for state in self._states:
+            if state.device_uuid == old_uuid:
+                state.device_uuid = new_uuid
+                state.update_for_device()
+                performed_swap = True
+        return performed_swap
 
 
 @QtQml.QmlElement
--- a/gremlin/macro.py
+++ b/gremlin/macro.py
@@ -23,7 +23,7 @@ import functools
 import logging
 import time
 from threading import Event, Lock, Thread
-from typing import Tuple
+from typing import override, Tuple
 import uuid
 from xml.etree import ElementTree
 
@@ -378,6 +378,10 @@ class AbstractAction(ABC):
         node.set("type", type_name)
         return node
 
+    def swap_uuid(self, old_uuid: uuid.UUID, new_uuid: uuid.UUID) -> bool:
+        """Swaps occurrences of the old UUID with the new one for this action."""
+        return False
+
 
 class JoystickAction(AbstractAction):
 
@@ -494,6 +498,13 @@ class JoystickAction(AbstractAction):
             self.value = util.read_property(
                 node, "value", PropertyType.HatDirection
             )
+
+    @override
+    def swap_uuid(self, old_uuid: uuid.UUID, new_uuid: uuid.UUID) -> bool:
+        if self.device_guid == old_uuid:
+            self.device_guid = new_uuid
+            return True
+        return False
 
 
 class KeyAction(AbstractAction):
--- a/gremlin/swap_devices.py
+++ b/gremlin/swap_devices.py
@@ -54,6 +54,9 @@ class SwapDevicesResult:
     action_swaps: int
     input_swaps: int
 
+    def as_string(self) -> str:
+        return f"Swapped {self.action_swaps} actions and {self.input_swaps} inputs."
+
 
 def _swap_device_inputs(
     profile: profile.Profile,
--- a/gremlin/swap_devices.py
+++ b/gremlin/swap_devices.py
@@ -22,7 +22,8 @@ Performs a simple swap of two devices in the profile.
 import dataclasses
 import uuid
 
-from gremlin import device_initialization, profile
+import gremlin.profile
+from gremlin import device_initialization
 
 
 @dataclasses.dataclass
@@ -32,9 +33,9 @@ class ProfileDeviceInfo:
     num_bindings: int = 0
 
 
-def get_profile_devices(pr: profile.Profile) -> list[ProfileDeviceInfo]:
+def get_profile_devices(profile: gremlin.profile.Profile) -> list[ProfileDeviceInfo]:
     profile_devices = {}
-    for device_uuid, inputs in pr.inputs.items():
+    for device_uuid, inputs in profile.inputs.items():
         if device_uuid in profile_devices:
             profile_devices[device_uuid].num_bindings += len(inputs)
         else:
@@ -53,13 +54,17 @@ def get_profile_devices(pr: profile.Profile) -> list[ProfileDeviceInfo]:
 class SwapDevicesResult:
     action_swaps: int
     input_swaps: int
+    user_script_swaps: int
 
     def as_string(self) -> str:
-        return f"Swapped {self.action_swaps} actions and {self.input_swaps} inputs."
+        return (
+            f"Swapped {self.action_swaps} actions, "
+            f"{self.input_swaps} inputs, and {self.user_script_swaps} user script vars."
+        )
 
 
 def _swap_device_inputs(
-    profile: profile.Profile,
+    profile: gremlin.profile.Profile,
     source_device_uuid: uuid.UUID,
     target_device_uuid: uuid.UUID,
 ) -> int:
@@ -81,7 +86,7 @@ def _swap_device_inputs(
 
 
 def _swap_device_actions(
-    profile: profile.Profile,
+    profile: gremlin.profile.Profile,
     source_device_uuid: uuid.UUID,
     target_device_uuid: uuid.UUID,
 ) -> int:
@@ -91,8 +96,19 @@ def _swap_device_actions(
     return swap_count
 
 
+def _swap_device_user_script_vars(
+    profile: gremlin.profile.Profile,
+    source_device_uuid: uuid.UUID,
+    target_device_uuid: uuid.UUID,
+) -> int:
+    swap_count = 0
+    for script in profile.scripts.scripts:
+        swap_count += script.swap_uuid(source_device_uuid, target_device_uuid)
+    return swap_count
+
+
 def swap_devices(
-    profile: profile.Profile,
+    profile: gremlin.profile.Profile,
     source_device_uuid: uuid.UUID,
     target_device_uuid: uuid.UUID,
 ) -> SwapDevicesResult:
@@ -111,4 +127,5 @@ def swap_devices(
     return SwapDevicesResult(
         _swap_device_actions(profile, source_device_uuid, target_device_uuid),
         _swap_device_inputs(profile, source_device_uuid, target_device_uuid),
+        _swap_device_user_script_vars(profile, source_device_uuid, target_device_uuid),
     )
--- a/gremlin/swap_devices.py
+++ b/gremlin/swap_devices.py
@@ -1,0 +1,111 @@
+# -*- coding: utf-8; -*-
+
+# Copyright (C) 2015 - 2025 Lionel Ott
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Performs a simple swap of two devices in the profile.
+"""
+
+import dataclasses
+import uuid
+
+from gremlin import device_initialization, profile
+
+
+@dataclasses.dataclass
+class ProfileDeviceInfo:
+    device_uuid: uuid.UUID
+    name: str = ""
+    num_bindings: int = 0
+
+
+def get_profile_devices(pr: profile.Profile) -> list[ProfileDeviceInfo]:
+    profile_devices = {}
+    for device_uuid, inputs in pr.inputs.items():
+        if device_uuid in profile_devices:
+            profile_devices[device_uuid].num_bindings += len(inputs)
+        else:
+            profile_devices[device_uuid] = ProfileDeviceInfo(
+                device_uuid, num_bindings=len(inputs)
+            )
+
+    # We can get names for connected devices (only).
+    for d in device_initialization.physical_devices():
+        if d.device_guid.uuid in profile_devices:
+            profile_devices[d.device_guid.uuid].name = d.name
+    return list(profile_devices.values())
+
+
+@dataclasses.dataclass
+class SwapDevicesResult:
+    action_swaps: int
+    input_swaps: int
+
+
+def _swap_device_inputs(
+    profile: profile.Profile,
+    source_device_uuid: uuid.UUID,
+    target_device_uuid: uuid.UUID,
+) -> int:
+    swap_count = 0
+    source_device_inputs = profile.inputs.get(source_device_uuid)
+    target_device_inputs = profile.inputs.get(target_device_uuid)
+
+    if source_device_inputs:
+        for input_item in source_device_inputs:
+            input_item.device_id = target_device_uuid
+            swap_count += 1
+        profile.inputs[target_device_uuid] = source_device_inputs
+    if target_device_inputs:
+        for input_item in target_device_inputs:
+            input_item.device_id = source_device_uuid
+            swap_count += 1
+        profile.inputs[source_device_uuid] = target_device_inputs
+    return swap_count
+
+
+def _swap_device_actions(
+    profile: profile.Profile,
+    source_device_uuid: uuid.UUID,
+    target_device_uuid: uuid.UUID,
+) -> int:
+    swap_count = 0
+    for action in profile.library.actions_by_predicate(lambda x: True):
+        swap_count += int(action.swap_uuid(source_device_uuid, target_device_uuid))
+    return swap_count
+
+
+def swap_devices(
+    profile: profile.Profile,
+    source_device_uuid: uuid.UUID,
+    target_device_uuid: uuid.UUID,
+) -> SwapDevicesResult:
+    """Swaps two devices in the profile (from a device in the profile to a connected device).
+
+    It is the caller's responsibility to ensure that the devices UUIDs are valid.
+
+    Args:
+        profile: The Profile to perform the swap on.
+        source_device_uuid: The UUID of the source (from profile) device.
+        target_device_uuid: The UUID of the target (connected) device.
+
+    Returns:
+        The SwapDevicesResult object containing stats on the swaps performed.
+    """
+    return SwapDevicesResult(
+        _swap_device_actions(profile, source_device_uuid, target_device_uuid),
+        _swap_device_inputs(profile, source_device_uuid, target_device_uuid),
+    )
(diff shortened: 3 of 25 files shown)


action_plugins/condition/condition.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABCMeta, abstractmethod
-from typing import Any, List, Optional, TYPE_CHECKING
+from typing import Any, List, Optional, override, TYPE_CHECKING
```
WhiteMagic, line 4: There's a ton of places in the whole action stuff that can use this annotation, so good to start somewhere 👍

gremlin/ui/swap_devices.py
```diff
@@ -0,0 +1,98 @@
...
+        if len(self._devices) == 0:
+            return str(dill.UUID_Invalid)
+        if not (0 <= index < len(self._devices)):
+            raise GremlinError("Provided index out of range")
+
+        return str(self._devices[index].device_uuid)
+
+
+@QtQml.QmlElement
+class SwapDevices(QtCore.QObject):
```
WhiteMagic: Not sure I understand the point of this. a QObject doesn't have anything useful for visualization purposes, but it's being exposed as an element? It also only exposes a single slot so that seems a bit overkill. To me this sounds like this would much better live in the class just above or in the backend, as it does nothing more than forward to some logic code.

gremlin/ui/profile_devices_model.py
```diff
@@ -0,0 +1,98 @@
...
+    def update_model(self) -> None:
+        """Updates the model if the connected devices change."""
+        self._devices = swap_devices.get_profile_devices(backend.Backend().profile)
+        # Ensure the entire model is refreshed
+        self.modelReset.emit()
+
+    def rowCount(self, parent: QtCore.QModelIndex = ...) -> int:
+        return len(self._devices)
+
+    def data(
```
WhiteMagic, line 67: I try not doing presentation logic in models, as that's the main point of the whole model-view-controller pattern. So it would be better to expose name, uuid, number of bindings, and whatever else is interesting as separate fields that the presentation QML side can put together as it sees fit. That makes it possible to redo the UI without having to modify the model code.

gremlin/macro.py
```diff
@@ -378,6 +378,10 @@ def _create_node(self, type_name: str) -> ElementTree.Element:
         node.set("type", type_name)
         return node
 
+    def swap_uuid(self, old_uuid: uuid.UUID, new_uuid: uuid.UUID) -> bool:
```
WhiteMagic, line 381: I'm not entirely sure I like having device id swap logic in the macro as this has nothing really to do with it. I see the convenience of doing it that way, however. I'll have to think on that a bit.

gremlin/swap_devices.py
```diff
@@ -0,0 +1,114 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Performs a simple swap of two devices in the profile.
+"""
+
+import dataclasses
+import uuid
+
+from gremlin import device_initialization, profile
```
WhiteMagic: Be careful with importing `profile` in this way. I have on more than one occasion got bitten by having a local `profile` variable that then got removed and I assigned bad things to `profile` and things broke in random unrelated places.

gremlin/swap_devices.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+@dataclasses.dataclass
+class ProfileDeviceInfo:
+    device_uuid: uuid.UUID
+    name: str = ""
+    num_bindings: int = 0
+
+
+def get_profile_devices(pr: profile.Profile) -> list[ProfileDeviceInfo]:
```
WhiteMagic: `pr` isn't a great variable name honestly, I try to keep them from being overly terse. I know the reason which is the kind of bad name of `profile` module :-) In most cases I really just do `import gremlin.profile` and live with that. Or if I know which classes I end up using I import those directly to avoid the `profile` variable name used up.

gremlin/swap_devices.py
```diff
@@ -0,0 +1,114 @@
...
+    profile_devices = {}
+    for device_uuid, inputs in pr.inputs.items():
+        if device_uuid in profile_devices:
+            profile_devices[device_uuid].num_bindings += len(inputs)
+        else:
+            profile_devices[device_uuid] = ProfileDeviceInfo(
+                device_uuid, num_bindings=len(inputs)
+            )
+
+    # We can get names for connected devices (only).
```
WhiteMagic, line 46: Yeah that's kind of a shame :-) There is the option of having a small additional thing in the profile XML that's say a `<device-database>...</device-database>` that does the mapping from uuid to a bunch of info. Not going to be overly useful a lot of the times but might be nice sometimes, like here.

gremlin/swap_devices.py
```diff
@@ -0,0 +1,114 @@
...
+            profile_devices[d.device_guid.uuid].name = d.name
+    return list(profile_devices.values())
+
+
+@dataclasses.dataclass
+class SwapDevicesResult:
+    action_swaps: int
+    input_swaps: int
+
+    def as_string(self) -> str:
```
WhiteMagic, line 59: If you want a method returning a string it should just be `def __str__(self) -> str` which is what Python will automatically use on str() or repr() if `__repr__` is missing.

qml/DialogSwapDevices.qml
```diff
@@ -0,0 +1,148 @@
...
+    ProfileDeviceListModel {
+        id: profileDevices
+    }
+
+    SwapDevices {
+        id: swapDevices
+    }
+
+    // Properties to track the selected devices and user selections.
+    property var selectedTargetDevice: ""
```
WhiteMagic: With the model view stuff you'd want to track that information internally inside the model, i.e. it knows which drop down elements are selected. The vJoy selector does that for example (with more complicated stuff thrown in).

qml/DialogSwapDevices.qml
```diff
@@ -0,0 +1,148 @@
...
+                    Component.onCompleted: {
+                        selectedTargetDevice = physicalDevices.guidAtIndex(0) || ""
+                    }
+                }
+            }
+
+            // Second ("Profile") Device Column
+            ColumnLayout {
+                spacing: 5
+                Layout.preferredWidth: parent.width / 2 - 10
```
WhiteMagic: Here instead of dealing with the spacing and offset parts you could also use the `LayoutHorizontalSpacer` I wrote which effectively is a rectangle with a fillwidth: true layout property. If places between these two columns it will just push things apart if you give it enough space.

qml/DialogSwapDevices.qml
```diff
@@ -0,0 +1,148 @@
...
+            spacing: 10
+
+            Button {
+                id: _swapButton
+                text: qsTr("Swap")
+                Layout.preferredWidth: implicitWidth + 20
+                Layout.preferredHeight: 30
+                onClicked: {
+                    console.log("Selected first device: " + selectedTargetDevice);
+                    console.log("Selected second device: " + selectedSourceDevice);
```
WhiteMagic: Once we know this works I'd kick these out. I tend to use the qml console log stuff only for debugging stuff

action_plugins/condition/condition.py
code-monet: I will make a separate PR adding these in more files.

gremlin/ui/swap_devices.py
code-monet: We had a similar discussion as with automapper, right? We _could_ put it into the backend, but then the latter gets heavier and heavier. Instead we could keep separate elements for these things and one day remove them from the backend as well. Let me know your preference for this PR, though.

gremlin/ui/profile_devices_model.py
code-monet: Moved to QML.

gremlin/swap_devices.py
code-monet: Good catch, changed. I need to enable warnings in my IDE for these things (if I knew where/how...)

gremlin/swap_devices.py
code-monet: Replaced with profile and changed import. Sometimes I also add an underscore to avoid collisions e.g. `profile_: profile.Profile`

gremlin/swap_devices.py
code-monet: In this case I'd not override `__str__` as it's useful for our own development and debugging.

qml/DialogSwapDevices.qml
code-monet: Agreed and removed.

qml/DialogSwapDevices.qml
code-monet: Actually I removed some of these specifications (that were generated by AI) and the UI still looks reasonable, just hadn't uploaded the commit. Good to know about `LayoutHorizontalSpacer`, would be handy elsewhere.

qml/DialogSwapDevices.qml
code-monet: Done for both models.

gremlin/swap_devices.py
code-monet: ...it feels like more than appropriate work to "be nice" :) but we can do it. Let me look into that tomorrow.

code-monet:
I could use your help with improving this bit of code in the QML:

```
            // Second ("Profile") Device Column
            ColumnLayout {
                Label {
                    text: qsTr("Select device in profile")
                    font.bold: true
                }
                
                ComboBox {
                    model: profileDevices
                    textRole: "uuid"
                    Layout.fillWidth: true
                    onActivated: profileDevices.selectedIndex = currentIndex

                    delegate: ItemDelegate {
                        width: parent.width
                        text: `${model.name || model.uuid}: ${model.numBindings} ${model.numBindings === 1 ? 'binding' : 'bindings'}`
                    }
                }
            }
```

I asked AI if it had ideas on how to use the formatted string from delegate for the textRole as well (so the selected value and drop down value are the same), but it suggested something rather elaborate.

gremlin/macro.py
WhiteMagic: After some pondering, since they already also support `to_xml()` and `from_xml()` methods which don't really serve an actual macro functionality purpose it makes sense to me to also have this ability in there. Every other option would just lead to headache later on.

gremlin/ui/swap_devices.py
WhiteMagic: Hmmm that's true, I think my initial instinct is that classes that are there just for one method are a bit odd. At the same time dumping all of this into the backend is not super nice. Mainly because the backend is meant to hold state, thus stateless aspects kind of don't need to be in there.

What do you think of having all these "tools" that do mapping parts etc. in one single class. The calls all don't require state, effectively being free-standing functions but not because of Qt/QML. That way any such "do this thing which doesn't really fit inside of a model property bit but I don't require state management" can live in that class or a few classes if we get to the point of having more with different focus, but they all are part of the `Gremlin.Tools` QML package/library.

WhiteMagic:
> I could use your help with improving this bit of code in the QML:
> 
> ```
>             // Second ("Profile") Device Column
>             ColumnLayout {
>                 Label {
>                     text: qsTr("Select device in profile")
>                     font.bold: true
>                 }
>                 
>                 ComboBox {
>                     model: profileDevices
>                     textRole: "uuid"
>                     Layout.fillWidth: true
>                     onActivated: profileDevices.selectedIndex = currentIndex
> 
>                     delegate: ItemDelegate {
>                         width: parent.width
>                         text: `${model.name || model.uuid}: ${model.numBindings} ${model.numBindings === 1 ? 'binding' : 'bindings'}`
>                     }
>                 }
>             }
> ```
> 
> I asked AI if it had ideas on how to use the formatted string from delegate for the textRole as well (so the selected value and drop down value are the same), but it suggested something rather elaborate.

Hmm so `ComboBox` uses `textRole` and `valueRole` to get the values to use from the model. So I don't think there is a "nice" way of doing this as it expects to be given the data from the model. What is the reason you want the presentation of the delegate to match the model's name? For example if you use the `valueRole` to be the UUID you can use this with `indexOfValue` to select the correct drop down entry and the delegate then is whatever you want it to be or look like.

gremlin/macro.py
code-monet: Sounds good, leaving as-is.

gremlin/ui/swap_devices.py
code-monet: Sounds good, created a new file and class for tools and moved both swap devices and auto mapper to it.

gremlin/swap_devices.py
code-monet: Gonna punt on this unless you would like to see it in this PR.

code-monet:
> What is the reason you want the presentation of the delegate to match the model's name?

Right now, the drop down when collapsed shows the `textRole` i.e. the UUID, but expanding the dropdown shows the presentation of the delegate. It would be nice to have the collapsed drop down have the same text, but not necessary (I'm avoiding having a `label` in the QML since you recommended going the presentation via QML route instead).

gremlin/swap_devices.py
WhiteMagic: Fine with me, I'll put it on the list of things to add just so it's not forgotten about. It's also a separate piece of work so would be better to be it's own MR etc anyways.

code-monet:
Rebased and fixed conflicts - let me know if you would like to see more changes, or please go ahead with merge otherwise.

gremlin/swap_devices.py
code-monet: Tracked in #657

action_plugins/condition/condition.py
```diff
@@ -491,6 +495,9 @@ def __init__(
             self.input_type = input_type
             self.input_id = input_id
             self.joystick = None
+            self.update_for_device()
+        
+        def update_for_device(self):
```
WhiteMagic: The method name isn't the most self explanatory with regards to what it's meant to do based solely on the name. Something like `lookup_from_uuid` or `initialize_from_uuid` could be a bit clearer, IMO.

Possibly might also be worth modifying the signature to `initialize_from_uuid(self, device_uuid)` that way it's clear where the uuid comes from and we don't run the risk of having this implied uuid flying around and it's always that function's job to save the uuid and attempt to retrieve the joystick instance.

gremlin/ui/device.py
```diff
@@ -184,6 +185,16 @@ def _change_device_type(self, types: str) -> None:
...
         self.rowsRemoved.emit(self.parent(), 0, new_count)
         self.rowsInserted.emit(self.parent(), 0, new_count)
+    
+    @QtCore.Property(int)
+    def selectedIndex(self):
+        return self._selected_index
+        
+    @selectedIndex.setter
+    def selectedIndex(self, index):
+        if 0 <= index < len(self._devices):
```
WhiteMagic: Often there is an additional check on the index not being the same as the currently selected one to avoid event loops. I'm not sure if this I something that can happen here.

gremlin/ui/profile_devices_model.py
```diff
@@ -0,0 +1,97 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from typing import Any
+
+from PySide6 import QtCore
```
WhiteMagic: I usually import all the modules in one line from the same package. Though I'm not sure what I like better :-) Benefit of one line per module is it's easy to see the diff, disadvantage is that it can get quite long. No need to change but also happy for opinion.

gremlin/ui/profile_devices_model.py
```diff
@@ -0,0 +1,97 @@
...
+    """Model listing devices with bindings in the profile."""
+    selectedIndexChanged = QtCore.Signal()
+
+    roles = {
+        QtCore.Qt.UserRole + 1: QtCore.QByteArray("name".encode()),
+        QtCore.Qt.UserRole + 2: QtCore.QByteArray("uuid".encode()),
+        QtCore.Qt.UserRole + 3: QtCore.QByteArray("numBindings".encode())
+    }
+
+    def __init__(self, parent=None):
```
WhiteMagic, line 51: There is a `gremlin.ui.type_annotations` file which contains a few type alias for the types that these Qt model functions supposedly use. Makes it easier to use the correct type annotation and should they change/be wrong it's easy enough to change.

gremlin/ui/profile_devices_model.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def data(
+        self, index: QtCore.QModelIndex, role: int = QtCore.Qt.DisplayRole
+    ) -> Any:
+        if not index.isValid() or not (0 <= index.row() < len(self._devices)):
+            return None
+
+        device = self._devices[index.row()]
+        if role in ProfileDeviceListModel.roles:
+            role_name = ProfileDeviceListModel.roles[role].data().decode()
```
WhiteMagic: I use this pattern in some older code but in newer one I figured out you can directly use the `role` without getting the string out of it. An example is here https://github.com/WhiteMagic/JoystickGremlin/blob/49f4b62e3c4663137108bcd3e65e7d7c5f2b5cbd/gremlin/ui/option.py#L41 it does the same thing but is a bit cleaner since it doesn't have to deal with the whole decode stuff.

gremlin/ui/profile_devices_model.py
```diff
@@ -0,0 +1,97 @@
...
+    @QtCore.Slot(int, result=str)
+    def uuidAtIndex(self, index: int) -> str:
+        if len(self._devices) == 0:
+            return str(dill.UUID_Invalid)
+        if not (0 <= index < len(self._devices)):
+            raise GremlinError(f"Provided {index=} out of range")
+
+        return str(self._devices[index].device_uuid)
+    
+    @QtCore.Property(int)
```
WhiteMagic: Are you positive this makes Qt happy enough? Normally, it complains when a property is defined without being told about the signal associated with it. Typically it  complains that this property is not notifable/notifying yet it is not declared constant (or something to that effect). That's why I often just use the Property function rather than the decorator.

gremlin/swap_devices.py
WhiteMagic: Ah ok you want this as the actual user facing info and not internal stringification. Technically that's one of the "better give QML the data so it can render the text as desired". Though I think that's quite the overhead for this case so I'm fine leaving it as is until we run into that as a need.

qml/DialogSwapDevices.qml
```diff
@@ -0,0 +1,133 @@
...
+
+
+Window {
+    minimumWidth: 900
+    minimumHeight: 300
+
+    title: "Swap devices: Swaps bindings between two devices"
+
+    DeviceListModel {
+        id: physicalDevices
```
WhiteMagic: I try to prefix all id's in QML code with `_` such that I can easily identify what bits are ids and what is just some other thing.

qml/DialogSwapDevices.qml
```diff
@@ -0,0 +1,133 @@
...
+
+            // First ("Connected") Device Column
+            ColumnLayout {
+                Label {
+                    text: qsTr("Select connected device")
+                    font.bold: true
+                }
+                
+                ComboBox {
+                    id: firstDeviceCombo
```
WhiteMagic: In general unless the id is needed for some interaction it's fine to omit it.

WhiteMagic:
> Rebased and fixed conflicts - let me know if you would like to see more changes, or please go ahead with merge otherwise.

Added a few comments here and there, nothing major. In general really nice and I appreciate you constantly adding tests as well. I'm at the moment (in the little bit of time I have right now) trying to find a way to beat Qt into submission to let me do actual "interaction tests" with regards to actions. The goal is to be able to have tests for the complex behaviours actions can have when nested etc.

gremlin/swap_devices.py
code-monet: Sounds good re: overhead, marking as resolved for this PR

gremlin/ui/profile_devices_model.py
code-monet: This must have been how AI did it; I'd normally do it as so:

```
from PySide6 import (
    QtCore,
    QtQml,
)
```

so that:
1. diffs are cleaner (as is blame)
2. the `black` autoformatter will fight and insist on doing it this way once the line gets long enough (I currently go back and revert it if it does that, per your preference).

I've done it all on one line for this PR.

---

I'm only realizing now that Ruff is a formatter as well - is that what you use to format code? I'll configure my IDE to do the same for this project.

gremlin/ui/profile_devices_model.py
code-monet: You mean `type_aliases.py`? Good to know, using that here now.

gremlin/ui/profile_devices_model.py
code-monet: Good to know, doing it that way now.

gremlin/ui/profile_devices_model.py
code-monet: I'm not seeing a warning in the terminal. Per AI, this is okay because the property setter is what is emitting the signal; a read-only property does not *need* an associated signal.

qml/DialogSwapDevices.qml
code-monet: Made the change and will try to remember for future.

qml/DialogSwapDevices.qml
code-monet: I just realized these were here - removed.

action_plugins/condition/condition.py
code-monet: Did both - for the latter, I now set `self.device_uuid` in this function as well to prevent the possibility of having a mismatch in the stored UUID and the one in the argument.

gremlin/ui/device.py
code-monet: Added the check. I think I've seen warnings in the past in the terminal where Qt says it stopped an infinite loop after 2 or so passes.

code-monet:
> I appreciate you constantly adding tests as well. I'm at the moment (in the little bit of time I have right now) trying to find a way to beat Qt into submission to let me do actual "interaction tests" with regards to actions. The goal is to be able to have tests for the complex behaviours actions can have when nested etc.

Of course (it's just in my DNA as a coder), and I appreciate you acknowledging that! Feel free to open issues for improving test coverage and assigning them to me/splitting them between us as well, I was planning to get back to that eventually.

gremlin/ui/profile_devices_model.py
WhiteMagic: I would say I'm fine with that style. It uses more lines but has obvious benefits. I wouldn't go and change them everywhere in one feel swoop, just whenever we have to touch imports anyways we can change to this style.

I don't use any tool to format things, I likely would have to spend quite a bit of time fighting it to adhere to the weird idiosyncrasies I do and/or I may be unable to make it split code the way I prefer and might get annoyed :-) But I've also not tried in a while

gremlin/ui/profile_devices_model.py
WhiteMagic: Hmm, not sure I trust the LLM on that. If it was a read-only property and marked up as such yes, no signal is needed. However, this property is not read-only as there is a setter as well.

Based on this https://doc.qt.io/qtforpython-6/PySide6/QtCore/Property.html (all the way at the bottom) you have to provide the signal that is used to notify about a change in property (this is due to the property binding that QML does) unless the property is defined as constant. One doesn't have to use the function form it's also possible to do that with the decorator syntax, e.g. `@Property(int, notify=changed)`

gremlin/ui/profile_devices_model.py
code-monet: Gotcha, I'll keep using the "Format modified lines" option in my IDE then. Good to be aligned on the multiline imports!

gremlin/ui/profile_devices_model.py
code-monet: Thanks for the reference - changed here and in `devices.py`. I challenged the AI on it and it insisted its suggestion was right :(

I think I don't need to manually emit the signal anymore (I would have to if `_selected_index` was being changed outside of the property setter...).

WhiteMagic: Heh LLM's once they're deep in their context just repeat themselves to some extent or at least do after three or so iterations of back and forth :-/.

Regarding the not having to emit the signal in the setter. I'm unsure if that's correct, everything I've ever run across emits those signals as this is what makes the whole machinery run. If it wasn't emitted PySide would have to behind the scenes assume that you want to emit that particular signal.

code-monet: You're right; from the [docs](https://doc.qt.io/qt-6/properties.html): 

> The signal is emitted automatically when the property is changed via the Qt API ([QObject::setProperty](https://doc.qt.io/qt-6/qobject.html#setProperty), [QMetaProperty](https://doc.qt.io/qt-6/qmetaproperty.html), etc.), but not when the MEMBER is changed directly.

So this is *not* a situation where the signal is emitted automatically.. Made the change.

WhiteMagic:
There may be things here and there that need modifying etc. based on usage but as a first version it looks good to me.

PR #656, opened 2025-11-27
Add @override decorator to action plugins

code-monet:
[Commit made by Jules bot. I eyeballed the changes and they look good. Pyright in my IDE has no issues with the type hints either]

Applied the @override decorator to methods in action plugins that override base class methods, specifically for subclasses of AbstractActionData and AbstractFunctor.

- Added `from typing import override`
- Decorated methods like `__call__`, `_from_xml`, `_to_xml`, `is_valid`, `_valid_selectors`, `_get_container`, `_handle_behavior_change` and others where appropriate.


WhiteMagic:
Had a look over it and I couldn't see any incorrect methods being overriden.

PR #658, opened 2025-11-28
Add pyright configuration to the TOML file

code-monet:
If you have Pyright set up in the IDE (seems to be default in Visual Studio Code), then this should show type warnings in the `Problems` tab. This way we can fix forward issues it identifies.

I took a stab at fixing one of the files. The current Pyright configuration is a bit relaxed; we can make it more aggressive if you like (the `Problems` tab might fill up a bit more :)

### DIFF
--- a/gremlin/base_classes.py
+++ b/gremlin/base_classes.py
@@ -17,11 +17,11 @@
 
 from __future__ import annotations
 
-import typing
 from abc import abstractmethod, ABC
 import copy
 import time
-from typing import Any, List, Tuple, Optional
+import typing
+from typing import Any, List, Self, Tuple, Type, Optional
 import uuid
 from xml.etree import ElementTree
 
@@ -84,11 +84,11 @@ class AbstractActionData(ABC):
     # Fields expected to be present in every action plugin. These define
     # information required by the plugin manager as well as UI and behavior
     # logic.
-    version : int = -1
-    name : str = ""
-    tag : str = ""
-    icon : str = ""
-    functor = None
+    version : int
+    name : str
+    tag : str
+    icon : str
+    functor: Type[AbstractFunctor]
     model = None
     properties: Tuple[ActionProperty, ...] = ()
     input_types: Tuple[InputType, ...] = ()
@@ -216,13 +216,13 @@ class AbstractActionData(ABC):
         )
         self._from_xml(node, library)
 
-    def to_xml(self) -> ElementTree.Element:
+    def to_xml(self) -> Optional[ElementTree.Element]:
         """Returns an XML node representing the instance's contents.
 
         Calls the implementation specific serialization routine.
 
         Returns:
-            XML node containing the instance's contents
+            XML node containing the instance's contents, or None if invalid
         """
         if not self.is_valid():
             print(f"This node is invalid {self.id}")
@@ -336,7 +336,7 @@ class AbstractActionData(ABC):
         else:
             if not (0 <= anchor <= len(container)):
                 raise GremlinError(
-                    f"{self.name}: specified anchor index '{anchor.id}' is " +
+                    f"{self.name}: specified anchor index '{anchor}' is " +
                     f"not present in container '{selector}'"
                 )
             if mode == DataInsertionMode.Append:
@@ -393,7 +393,7 @@ class AbstractActionData(ABC):
             cls,
             mode: DataCreationMode,
             behavior_type: InputType
-    ) -> AbstractActionData:
+    ) -> Self:
         """Allows customization of instance creation in derived classes.
 
         Args:
@@ -403,7 +403,7 @@ class AbstractActionData(ABC):
         Returns:
             Newly created instance
         """
-        pass
+        raise NotImplementedError
 
     @abstractmethod
     def _valid_selectors(self) -> List[str]:
--- a/gremlin/base_classes.py
+++ b/gremlin/base_classes.py
@@ -20,18 +20,17 @@ from __future__ import annotations
 from abc import abstractmethod, ABC
 import copy
 import time
-import typing
-from typing import Any, List, Self, Tuple, Type, Optional
+from typing import Any, List, Self, Tuple, Type, TYPE_CHECKING, Optional
 import uuid
 from xml.etree import ElementTree
 
 from gremlin import util, event_handler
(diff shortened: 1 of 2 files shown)


gremlin/base_classes.py
```diff
@@ -17,11 +17,11 @@
 
 from __future__ import annotations
 
-import typing
 from abc import abstractmethod, ABC
 import copy
 import time
-from typing import Any, List, Tuple, Optional
+import typing
+from typing import Any, List, Self, Tuple, Type, Optional
```
WhiteMagic: This should not import typing if it already imports specific parts from the same module.

gremlin/base_classes.py
```diff
@@ -84,11 +84,11 @@ class AbstractActionData(ABC):
...
-    version : int = -1
-    name : str = ""
-    tag : str = ""
-    icon : str = ""
-    functor = None
+    version : int
+    name : str
+    tag : str
+    icon : str
+    functor: Type[AbstractFunctor]
```
WhiteMagic, line 90: May need to think of whether having defaults is desirable.

gremlin/base_classes.py
```diff
@@ -393,7 +393,7 @@ def _do_create(
             cls,
             mode: DataCreationMode,
             behavior_type: InputType
-    ) -> AbstractActionData:
+    ) -> Self:
```
WhiteMagic, line 395: What's the reason pyright gives for wanting self rather than the clearer abstractactiondata?

gremlin/base_classes.py
```diff
@@ -403,7 +403,7 @@ def _do_create(
         Returns:
             Newly created instance
         """
-        pass
+        raise NotImplementedError
```
WhiteMagic: This is wrong. Derived classes are perfectly fine to not override this. This is a "may override" but not a "have to implement" as is the case with XML parsing. As this class uses the ABC meta class, enforcement of implementation is indicated by the `@abstractmethod`. Also in case of a missing implementation needing to be enforced via an exception there's a specific exception in `gremlin.error`.

pyproject.toml
```diff
@@ -28,4 +28,13 @@ build-backend = "poetry.core.masonry.api"
...
+lint.ignore = ["ANN401"]
+
+[tool.pyright]
+include = ["action_plugins", "gremlin", "joystick_gremlin.py"]
+exclude = ["**/__pycache__", "**/.git", "**/.venv", "**/venv"]
+reportMissingImports = true
+reportMissingTypeStubs = false
+pythonVersion = "3.12"
+pythonPlatform = "Windows"
+typeCheckingMode = "basic"
```
WhiteMagic: Does specifying `include` turn this into a "whitelist only" mode? If so that's good. We can also add `dill` and `vjoy` folders as they contain Python code. What's the impact of specifying the `pythonVersion` and is it possible to specify multiple version?

gremlin/base_classes.py
code-monet: Importing TYPE_CHECKING as well now.

gremlin/base_classes.py
code-monet: Defaults that are placeholder values is useful if we check for those - if we're not checking, then better to not have defaults, that might give a more obscure error, possibly later in code execution. Just my 2c.

gremlin/base_classes.py
code-monet: This is not Pyright but me :) this is a [common use case](https://docs.python.org/3/library/typing.html#typing.Self) for `Self`:

> Other common use cases include: [classmethod](https://docs.python.org/3/library/functions.html#classmethod)s that are used as alternative constructors and return instances of the cls parameter.

Using `Self` is a narrower specification, implying that an instance of the current (sub)class is returned. `AbstractActionData` is broader and would imply an instance of the parent class also (although in this that cannot be instantiated because it is abstract).

gremlin/base_classes.py
code-monet: Pyright doesn't like it: "Function with declared return type "Self@AbstractActionData" must return value on all code paths; Type "None" is not assignable to type "Self@AbstractActionData""

If this is a "may override", the default implementation should return something that matches the type hint. Do you want me to make the return an `Optional`?

pyproject.toml
code-monet: [Reference](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#sample-pyprojecttoml-file)

1. Correct, explicit include turns off the default behavior of "everything in the tree containing the TOMLfile" and instead uses a whitelist
2. Added `dill` and `vjoy`
3. If `pythonVersion` is not specified, it will use the version of the Python interpreter active/installed. Thinking more, we should probably also add all announced Python versions so we are a bit more ready when the time comes - I'll add .13 and .14.

WhiteMagic: I just looked at the doc for `pythonVersion` it seems that you can only provide a single version. So we should pick the lowest one that is used and then over time bump it up as there is a need for features from newer versions. So for now I think 3.12 is fine, despite me using both 3.12 and 3.13 depending where I am.

gremlin/base_classes.py
WhiteMagic: Ah ok, I read up in it's use a bit and it should possibly actually fix a few complaints pylance has as well.

gremlin/base_classes.py
WhiteMagic: I looked at the code more and we can use the exception, as calling this method without it having been overridden in the derived class is not a valid use case. Thus using the `gremlin.error.MissingImplementationError` with a description is fine.

gremlin/base_classes.py
WhiteMagic: Hmmm, yes could argue that if they're not there one might not use them for verification either :-) Though the reason I added them I think was to be able to quickly test if them being defined that way gets modified as intended. So not having defaults is fine here since it's not been used in the past anyway.

gremlin/base_classes.py
code-monet: Using that exception now.

pyproject.toml
code-monet: I should have looked at the doc more closely - fixed. Pyright was working regardless, I guess it used the default option value.

PR #660, opened 2025-11-29
Add a database of devices to the profile

### LINKED ISSUE #657: Implement per profile device database
Profiles can contain inputs from devices that are not connected to the computer. For such entries no information about the device that the input is assigned to is available. In some situations having this information, i.e. the content of the `dill.DeviceSummar`, would be nice for both presentation and logic ideas.

The idea is to add a new section to the profile XML like `<device-database>` which stores device information that can then be leveraged by other parts in Gremlin.

code-monet:
Fixes #657

### DIFF
--- a/gremlin/device_initialization.py
+++ b/gremlin/device_initialization.py
@@ -1,5 +1,7 @@
+import collections
 import logging
 import threading
+import uuid
 
 import dill
 from gremlin import error, util
@@ -7,7 +9,7 @@ from vjoy.vjoy import VJoyProxy
 from vjoy import vjoy
 
 
-_joystick_devices = []
+_joystick_devices: dict[uuid.UUID, dill.DeviceSummary] = collections.OrderedDict()
 _joystick_init_lock = threading.Lock()
 
 
@@ -152,8 +154,9 @@ def joystick_devices_initialization() -> None:
         [dev for dev in devices if dev.is_virtual],
         key=lambda x: x.vjoy_id
     ))
-
-    _joystick_devices = sorted_devices
+    _joystick_devices.clear()
+    for dev in sorted_devices:
+        _joystick_devices[dev.device_guid.uuid] = dev
     _joystick_init_lock.release()
 
 
@@ -163,7 +166,7 @@ def joystick_devices() -> list[dill.DeviceSummary]:
     Returns:
         List containing information about all joystick devices
     """
-    return _joystick_devices
+    return list(_joystick_devices.values())
 
 
 def vjoy_devices() -> list[dill.DeviceSummary]:
@@ -172,7 +175,7 @@ def vjoy_devices() -> list[dill.DeviceSummary]:
     Returns:
         List of vJoy devices
     """
-    return [dev for dev in _joystick_devices if dev.is_virtual]
+    return [dev for dev in _joystick_devices.values() if dev.is_virtual]
 
 
 def physical_devices() -> list[dill.DeviceSummary]:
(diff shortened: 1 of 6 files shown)


gremlin/device_initialization.py
```diff
@@ -142,18 +143,8 @@ def joystick_devices_initialization() -> None:
...
-        key=lambda x: x.name
-    )
-    sorted_devices.extend(sorted(
-        [dev for dev in devices if dev.is_virtual],
-        key=lambda x: x.vjoy_id
-    ))
-
-    _joystick_devices = sorted_devices
+    # are made.
+    _joystick_devices = devices
```
WhiteMagic: Removing this logic causes a regression in behavior. The ordering enforced here guarantees that devices are always presented in the same order in the UI. Removing the ordering means that for different users devices will appear in different orders. The vJoy aspect isn't important anymore with how R14 handles things but removing device name based ordering is not ideal.

Sorting them does not have to happen here, but could be "outsourced" to the `physical_devices()` call.

gremlin/profile.py
```diff
@@ -498,6 +497,83 @@ def _parse_xml_action(self, action: ElementTree.Element) -> None:
...
+        """
+        return cls(device_uuid=uuid.UUID(node.get("uuid")), name=node.get("name", ""))
+
+    def to_xml(self) -> ElementTree.Element:
+        """Returns an XML node representing this device information.
+
+        Returns:
+            XML node containing the device's information
+        """
+        node = ElementTree.Element(self.XML_TAG)
```
WhiteMagic, line 525: All other XML de/serialization code leverages various utilities from the `gremlin.utils` module. I've also adopted a somewhat verbose way of encoding information as you've probably seen from the profiles. This is done through these helpers and they enforce that the correct data type is saved and loaded. Kind of a "crappy" XML schema. The benefit is that the code has to state what types it wants and the serialization checks that this holds and vice versa also when loading to avoid getting complicated errors when loading the "wrong" data since Python doesn't care much about that.

gremlin/device_initialization.py
code-monet: Ah, I should have thought more about why you might be sorting it. Using `OrderedDict` now and reverted some of my edits.

gremlin/profile.py
code-monet: I noticed that previously but forgot about it :/ made the change. I figure `DeviceDatabase` is fine as is, but let me know if you had something else in mind for how the XML for that should be done.

gremlin/profile.py
```diff
@@ -498,6 +497,86 @@ def _parse_xml_action(self, action: ElementTree.Element) -> None:
...
+        Args:
+            node: XML node containing the device information
+
+        Returns:
+            New _DeviceInfo instance initialized with data from the XML node
+        """
+        return cls(
+            device_uuid=read_subelement(node, "device-id"),
+            name=read_subelement(node, "device-name"),
+        )
```
WhiteMagic: This behavior while sensible, is different from any other `from_xml` method that exists in Gremlin. They all are member methods not classmethods. So while technically this is probably nicer I'd keep it the same as all the others to avoid confusion as to why is different.

gremlin/profile.py
WhiteMagic: Yeah for this simple case, which is similar to the `<input>...</input>` entries I'm fine with this form.

WhiteMagic:
Just have that one comment otherwise looks good to me.

gremlin/profile.py
code-monet: Consistency has its upside - made the change.

code-monet:
Rebased on top of dev as well.

WhiteMagic:
Nice!

code-monet:
I'll make a PR for the broken test - and a fix for a logic error that got introduced in this PR.

PR #661, opened 2025-11-30
Action interaction testing framework

WhiteMagic:
Framework to test the interaction of action sets using the logical device. Leverages pytest-qt in order to be able to emit events correctly from other threads. This is required for testing things where for example a button is held down (i.e. released in the future) and another event happens. Setting this up as a test without having the `QApplication` exist is infeasible.

Profiles under test are normal XML files which leverage 4 input and 4 output instances of the logical device for each input type.


code-monet:
> This is required for testing things where for example a button is held down (i.e. released in the future) and another event happens.

Do I understand right, that this kind of testing can also be done with the `test/integration/action_plugins` framework *except* for events that are not input events? Or is it the direct nature of creating and emitting events that's desirable?

WhiteMagic:
Certain aspects the `integration/action_plugins` does support, especially sequential ones. One drawback of that system though is that since the `processEvents` is manually called to let Qt process events it will be impossible to handle scenarios where events get triggered in the background while the test is emitting other events of it's own. Effectively, reproducing certain types of user interactions will be impossible to test for due to Qt's event loop fun.

I'm currently still seeing how to best structure this as using manual timing is very cumbersome for some aspects. The pytest-qt framework has a notion of waiting for signals, which I'm looking at at the moment to see if it can be abused or something similar done to express things such as "I expect these events to occur in this order" for example.

PR #665, opened 2025-12-10
Fix bug where periodics from only one script were running.

code-monet:
I attempted a fix in #640 but that broke this other aspect of periodics

Ideally, a script reload would also clean up its own periodics, but that's a larger refactor - perhaps we'd need a `ScriptPeriodicRegistry`, analogous to `ScriptVariableRegistry`.

Tested by running a profile with two scripts that both have periodics - previously, periodics from only one script would run.

### DIFF
--- a/gremlin/code_runner.py
+++ b/gremlin/code_runner.py
@@ -454,6 +454,7 @@ class CodeRunner:
         # Retrieve the list of current paths searched by Python
         system_paths = [os.path.normcase(os.path.abspath(p)) for p in sys.path]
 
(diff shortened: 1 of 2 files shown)


WhiteMagic:
Ahhh I only had tests with single periodic since in the past the issue was them either not running or suddenly multiple instances of the same running.

WhiteMagic:
Merging this and added a new task for the smarter way of handling it so it's not forgotten about (sadly it's cursed though #666).

PR #684, opened 2026-01-18
feat(device_db): Add T-LCM Pedals device mapping

ClementCornut:
Tested this by directly modifying the json in the current release.

Naming can be discussed. I wanted it to be as general as possible.


WhiteMagic:
Theoretically one could also use clutch, brake, and accelerator but the positional ones are more direct so I'm fine with those.

Issue #704, opened 2026-01-30
[R14] Mouse Button Actions in Macro don't trigger


djscott00:
Triggering of mouse button events does not appear to be working correctly when they're done via macros:

<img width="1025" height="622" alt="Image" src="https://github.com/user-attachments/assets/30129808-ca73-4a6b-b0c7-d0dc0d42cd53" />

They do work just fine when done via the main "Map to Mouse" action.

Tested this using https://www.onlinemictest.com/mouse-test/
Tested on the "joystick_gremlin_nightly_2026_01_28" nightly.

djscott00:
Should be fixed now in 0a35bdc90382d3327d0e5cf4f0443650ca4077bb

PR #707, opened 2026-01-31
[R14] Fix for mouse actions not triggering in macros.

djscott00:
Fix for [R14] Mouse Button Actions in Macro don't trigger(https://github.com/WhiteMagic/JoystickGremlin/issues/704)

Just needed some simple changes to the importing to correctly reference sendinput


WhiteMagic:
Looks good to me, thanks.

PR #708, opened 2026-01-31
Fix e2e integration tests now that loading a profile deactivates gremlin and invalidates vjoy devices

code-monet:
[This commit](https://github.com/WhiteMagic/JoystickGremlin/commit/75f8212e60af6eb8fbe82313814fdb32fa9fc1f3#diff-391e5281c5486339e8fe5831cb86e525b79cf6a68e04f8c5c789a0013c7538dd) broke the tests

The fix is pretty simple, the tests were incorrectly instantiating vJoy devices early to populate the profile when the just needed the vJoy device IDs. Verified by running locally.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -53,7 +53,7 @@ def vjoy_di_device(
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


WhiteMagic:
Ah thanks for fixing them, I didn't manage to look at them when I saw they failed locally.

PR #737, opened 2026-03-16
Added extra warning to quickstart when using multiple vjoy devices.

djscott00:
Small update to quickstart documentation to include a warning about multiple vjoy devices


WhiteMagic:
Definitely can't hurt with how often people get this wrong.

PR #743, opened 2026-03-27
Limit pyside to a version where the input viewer still works. 

code-monet:
Also update deploy.bat to use the same commands as in nightly_build.yml

On PySide 6.11 the Input Viewer looks as follows:

image


WhiteMagic:
Closing this as I fixed the issue around PySide 6.11 - was the simply fact that a property didn't expose a get method only set and notification which seemingly was ok prior to 6.11 but resulted in state not being set thus failing to provide data.

The deploy.bat isn't really used locally, and likely actually I should change some parts. Specifically, commit the poetry.lock file to ensure consistent package versions are installed without any guesswork.

code-monet:
Nice, better to have the fix than the constraint!

PR #750, opened 2026-04-29
In SelectionVariable, handle attempts to set invalid incoming gracefully

code-monet:
Otherwise, if the profile-saved current index becomes too large for a SelectionVariable (I have a plugin where this is populated by detecting devices), we keep getting exceptions at various points.

### DIFF
--- a/gremlin/user_script.py
+++ b/gremlin/user_script.py
@@ -1004,9 +1004,22 @@ class SelectionVariable(AbstractVariable):
     ) -> None:
         super().__init__(name, description, is_optional)
 
+        if default_index >= len(option_list):
+            raise ValueError(
+                f"Default index {default_index} is out of range for option list "
+                f"of length {len(option_list)} for selection variable '{name}'"
+            )
         self._option_list = option_list
-        self._current_index = default_index
+        self._set_current_index(default_index)
         self._initialize_from_registry()
+    
+    def _set_current_index(self, index: int) -> None:
+        if index < len(self._option_list):
+            self._current_index = index
+        else:
+            logging.getLogger("user_script").error(
+                f"Ignoring invalid index {index} for selection variable '{self.name}'"
+            )
 
     @property
     def options(self) -> list[str]:
@@ -1024,9 +1037,9 @@ class SelectionVariable(AbstractVariable):
         return True
 
     def _from_xml(self, node: ElementTree.Element) -> None:
-        self._current_index = util.read_property(
+        self._set_current_index(util.read_property(
             node, "index", PropertyType.Int
-        )
+        ))
 
     def _to_xml(self, node: ElementTree.Element) -> None:
         node.append(util.create_property_node(
@@ -1034,7 +1047,7 @@ class SelectionVariable(AbstractVariable):
         ))
 
     def _assign_value_from(self, other: SelectionVariable) -> None:
-        self._current_index = other._current_index
+        self._set_current_index(other._current_index)
 
 
 class StringVariable(AbstractVariable):
--- a/gremlin/user_script.py
+++ b/gremlin/user_script.py
@@ -1004,8 +1004,8 @@ class SelectionVariable(AbstractVariable):
     ) -> None:
         super().__init__(name, description, is_optional)
 
-        if default_index >= len(option_list):
-            raise ValueError(
+        if not (0 <= default_index < len(option_list)):
+            raise error.PluginError(
                 f"Default index {default_index} is out of range for option list "
                 f"of length {len(option_list)} for selection variable '{name}'"
             )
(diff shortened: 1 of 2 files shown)


gremlin/user_script.py
```diff
@@ -1004,9 +1004,22 @@ def __init__(
     ) -> None:
         super().__init__(name, description, is_optional)
 
+        if default_index >= len(option_list):
```
WhiteMagic: If we're already adding safe guards it should be complete. In this case the user could also set the value to something negative. Thus guarding against both `< 0` and `>= len(option_list)` or turning it on it's head with `0 <= default_index < len(option_list)` would be ideal.

Could also debate whether raising an exception is the "right" thing to do. We could also simply set the index to 0 and log a warning. Doesn't break things, but still logs that something happened.

gremlin/user_script.py
```diff
@@ -1004,9 +1004,22 @@ def __init__(
...
+                f"Default index {default_index} is out of range for option list "
+                f"of length {len(option_list)} for selection variable '{name}'"
+            )
         self._option_list = option_list
-        self._current_index = default_index
+        self._set_current_index(default_index)
         self._initialize_from_registry()
+    
+    def _set_current_index(self, index: int) -> None:
+        if index < len(self._option_list):
```
WhiteMagic: Same comment regarding negative index value.

gremlin/user_script.py
```diff
@@ -1004,9 +1004,22 @@ def __init__(
     ) -> None:
         super().__init__(name, description, is_optional)
 
+        if default_index >= len(option_list):
+            raise ValueError(
```
WhiteMagic: Unsure if `ValueError` or `GremlinError` is better suited. Technically it's a value error but realistically it's not a runtime error but a code error and it's inside of Gremlin which usually means GremlinError is used.

There is also `error.PluginError` which is meant for user script errors (the exception needs renaming) so that could be the other good option.

gremlin/user_script.py
```diff
@@ -1004,9 +1004,22 @@ def __init__(
...
+        self._set_current_index(default_index)
         self._initialize_from_registry()
+    
+    def _set_current_index(self, index: int) -> None:
+        if index < len(self._option_list):
+            self._current_index = index
+        else:
+            logging.getLogger("user_script").error(
+                f"Ignoring invalid index {index} for selection variable '{self.name}'"
+            )
```
WhiteMagic: Gremlin's code is quite inconsistent in what error/warning/info mean and how they are used. So often this might be just a warning as we recover from it, but in reality it's an error we mitigated. So personally torn on what I'd use.

WhiteMagic:
Makes sense to guard and mitigate these. I didn't have "dynamic" selection lists in mind when designing this variable.

gremlin/user_script.py
code-monet: Changed to also check for negative index.
I'm of the opinion that the right thing is to raise rather than deal with it quietly, which could potentially lead to surprising behavior later in execution.

gremlin/user_script.py
code-monet: Done

gremlin/user_script.py
code-monet: Using error.PluginError

gremlin/user_script.py
code-monet: Switched to warning, seems more appropriate: "something wrong happened, but there's a mitigation that *might* work".

code-monet:
Changes made, please take a look.

As an aside, perhaps it would have been better to save the default selection value rather than index, allowing more robust checking of script changes vs what's saved in the profile. But perhaps not worth changing the profile schema for.

WhiteMagic:
Looks good, thanks.

> Changes made, please take a look.
> 
> As an aside, perhaps it would have been better to save the default selection value rather than index, allowing more robust checking of script changes vs what's saved in the profile. But perhaps not worth changing the profile schema for.

Ugh on that topic  yes, I agree. I just had this lengthy battle with the vJoy selection widget where exactly that was what broke everything, trying to keep selection indices in sync with what had been selected when devices can vanish etc. So in the end I rewrote it to track the content rather than index. Which on the surface feels wrong but is more reliable. I guess here you have the same issue with the dynamic part changing what index corresponds to the actual selection.

I'm not opposed to changing that implementation detail. It will break the profile definition, so I will have to figure out how to do the profile conversion as that would be the first time this is needed. We can also wait until I figure out a sane mechanism for profile transformation. The old versions aren't what I'd call sane, as it was me manually specifying how to transmute the XML file until they were at the latest version.

code-monet:
We can wait - what made me think of this was, with this PR we switch to the "default" if the index is invalid, but it may be possible that the default is also no longer what the user thinks it is. Since it is a bit of an edge case I'd say we can wait on this.

Yeah tracking content vs index seems less performant on the face of it, but maybe with modern CPUs it isn't so much worse!

WhiteMagic:
The CPU overhead doesn't really matter for this kind of UI work, it just feels that indices are cleaner but get quickly messy when you behind the scenes track text values :-)

PR #778, opened 2026-06-11
Add Light / Dark / High Contrast Dark colour modes with live switching

SubliminalsTV:
### Summary

Replaces the single `dark-mode` boolean with three selectable UI color modes, switchable **live** (no restart):

- **Light** — unchanged from the current light theme.
- **Dark** — a soft charcoal palette (`#1E1E1E` background / `#D4D4D4` foreground) that's easier on the eyes than pure black.
- **High Contrast Dark** — the original pure-black/white dark behavior, preserved exactly.

The mode is chosen from a `color-mode` Selection in the general config and applied through a live binding, so changing it updates the UI immediately.

### How it's wired (and why it's backward-compatible)

- `qml/Style.qml` gains a `colorMode` string property; `isDarkMode` is **kept** as a derived value (`colorMode !== "Light"`), so every existing consumer of `Style.isDarkMode` (Main.qml, ResponseCurveAction.qml, the new `theme/Gremlin/` files) keeps working untouched.
- `gremlin/ui/backend.py` adds a `colorMode` property and **keeps** `useDarkMode`, now derived from `color-mode != "Light"` — no QML/Python consumer of `useDarkMode` breaks.
- Supporting theming so the new modes render correctly: the action-icon pen color in `action_image_generator.py` follows the active mode, plus minor tab-bar polish (`DeviceTabBar.qml`, `JGTabButton.qml`).

### Config migration — one deliberate behavior change

This removes the `global/general/dark-mode` boolean config key and replaces it with `color-mode`. There is **no migration shim**, so a user who previously had `dark-mode: true` will fall back to **Light** on first launch after upgrading (the new key defaults to Light). I left it out to keep the change minimal, but I'm happy to add a one-time migration (read old `dark-mode`, seed `color-mode` to "High Contrast Dark") if you'd prefer existing dark-mode users land on a dark variant. Your call.

### Testing

Live mode-switching verified in a compiled build (all three modes, icons and tab bar included). Rebased cleanly onto current `develop`, including the recent `theme/Gremlin/` rework.

### DIFF
--- a/gremlin/ui/action_image_generator.py
+++ b/gremlin/ui/action_image_generator.py
@@ -14,6 +14,16 @@ from PySide6 import (
     QtQuick,
 )
 
+from gremlin.config import Configuration
+
+# Glyph pen colour (R, G, B) per UI colour mode. Mirrors the foreground colours
+# in qml/Style.qml so the icons match the rest of the theme.
+_PEN_COLORS = {
+    "Light": (0, 0, 0),
+    "Dark": (0xD4, 0xD4, 0xD4),
+    "High Contrast Dark": (255, 255, 255),
+}
+
 
 class ActionSummaryImageProvider(QtQuick.QQuickImageProvider):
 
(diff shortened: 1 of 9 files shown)


WhiteMagic:
Not against other "themes" but not convinced this approach is scalable or that the LLM didn't take shortcuts that mess things up or misunderstand actual intent. Will have to think about it, but likely will require an actual "theme" feature rather than going from 2 to 3, as that breaks the 0, 1, N rule.

SubliminalsTV:
My initial change was to fix a bug with the UI that made dark mode unusable for most people. But I figured I would remove the requirement to restart the app and also keep the current dark mode as "high contrast".

I do use an LLM but I do daily upkeep with its context to ensure that "take shortcuts that mess things up or misunderstand actual intent." stuff like what you mentioned doesn't happen.

I do have background in design and front end development. So I feel qualified to guide AI to do things I can't do on my own.

However, I never thought for second you would blindly add any changes so I will keep PRs short so you can look over it and decide for yourself if it should be added or you modify before adding.

WhiteMagic:
None of the changes are neccessarily problematic but as a single MR they don't work.
- I think the fix for the bug (I assume the color of action icons being black instead of white?) should be one MR.
- Then the live refresh which definitely is nice would be its own thing, as from a quick glance I'm not sure that's the way I'd do it but it's definitely a starting point to figure out what works best for Gremlin.
- The "colorscheme" support is definitely something interesting, but wouldn't work the way it's done now. It works for that intent, but I know how things go with Gremlin a "this works well for this idea" never survives when someone else sees it and says "oh you have X that's close to Y would be nice to have that". Which typically is true but because of how X was done Y is actually a lot of work. So that's why I try to think of the other ways something like "oh have a modified color scheme" could change in the future and then build support for that eventual use case in.
  - For a the color part it likely would be trying to let people define/save colors for certain Universal color names such that the UI can update properly. Though that's just a rough guess that hasn't been thought through fully.
  - Caveat there is now "theme" things in Gremlin but those aren't color themes they are just a way to have more compact UI elements instead of just the rather spacious default ones. 
- There seems also to be some unrelated reordering of widgets going on?

SubliminalsTV:
That breakdown makes sense — agreed it's too much for one MR. Closing this and splitting it the way you laid out.

The icon color bug — glyphs drawn black, invisible against the dark background — on its own. That's the actual bug that started this for me.
Live refresh as its own thing. It's more a starting point than a finished design; happy to rework it whatever way fits Gremlin best.
The color-scheme part I'll hold and open as an issue first instead of throwing code at you. Your point about building for the eventual "oh that's close to Y, can we have that too" case is the right instinct, and the user-defined Universal-color-name angle is worth thinking through before anyone writes a line. I'd rather get the shape right with you than hand you a design that only fits my one use case. Noted too that the existing "theme" stuff is compact-UI, not color — different thing.
And yeah, the widget reordering was unrelated and shouldn't have been in there. Pulled out.

Appreciate you taking the time to break it down instead of just closing it.

PR #781, opened 2026-06-14
Apply UI config changes live without requiring a restart

SubliminalsTV:
Config changes that affect the UI — dark mode being the obvious one — currently only take effect on restart. This emits the existing `configChanged` signal when a config entry is written and has `Main.qml` re-apply on it, so toggling dark mode updates the palette immediately.

Per your note on #778, treat this as a starting point rather than a finished design — it's the smallest thing that works. If there's a direction you'd rather Gremlin take for live config, I'm glad to redo it that way; I mostly wanted something concrete to react to.

Scoped to dark mode for now. Action-icon glyphs won't recolor live yet since they're cached — that's a separate follow-up, not in here.

### DIFF
--- a/gremlin/ui/option.py
+++ b/gremlin/ui/option.py
@@ -31,6 +31,7 @@ from gremlin.error import (
     GremlinError,
     MissingImplementationError,
 )
+from gremlin.signal import signal
 from gremlin.tts import TTSManager
 from gremlin.types import PropertyType
 
@@ -250,6 +251,8 @@ class ConfigEntryModel(QtCore.QAbstractListModel):
 
             self._config.set(*key, value)
             self.dataChanged.emit(index, index, {role})
+            # Enable other UI elements to react to configuration changes.
+            signal.configChanged.emit()
             return True
         return False
 
--- a/qml/Main.qml
+++ b/qml/Main.qml
@@ -31,6 +31,7 @@ ApplicationWindow {
     Component.onCompleted: () => {
         Style.isDarkMode = backend.useDarkMode
     }
+
     Universal.theme: Style.theme
     color: Style.background
 
@@ -426,6 +427,12 @@ ApplicationWindow {
     }
     Connections {
         target: signal
+
+        // Re-apply UI config (e.g. dark mode) live, so toggling it in the
+        // options takes effect immediately instead of needing a restart.
+        function onConfigChanged() {
+            Style.isDarkMode = backend.useDarkMode
+        }
 
         function onShowError(message, details) {
             _errorDialog.text = message
--- a/joystick_gremlin.py
+++ b/joystick_gremlin.py
@@ -168,7 +168,7 @@ def register_config_options() -> None:
     cfg.register(
         "global", "general", "dark-mode",
         PropertyType.Bool, False,
-        "Use the dark mode UI (requires restart).", {}, True
+        "Use the dark mode UI.", {}, True
     )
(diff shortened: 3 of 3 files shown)


SubliminalsTV:
No rush on this — just flagging it's ready for review whenever you have a window. It's file-disjoint from #780/#783, so there's no ordering dependency between them.

gremlin/ui/option.py
```diff
@@ -250,6 +251,9 @@ def setData(
 
             self._config.set(*key, value)
             self.dataChanged.emit(index, index, {role})
+            # Let UI elements that can update live (e.g. the dark-mode palette)
+            # react immediately rather than only on the next restart.
+            signal.configChanged.emit()
```
WhiteMagic: Similar to other comment related comments, I'd keep this more general, e.g. "Enable other UI elements to react to configuration changes." That should be clear enough as to why it's emitted without being specific to the one issue it addresses initially. Since there may other places this signal could be used, e.g. i) reordering of actions (currently updates when you change input for example) or ii) input item display mode (currently applies once you change the device tab).

WhiteMagic:
No big changes.

Also just as a heads up, in case you're thinking of complex future changes, I've played around with claude design to see what it could come up with as a refined UI style, as I personally don't like dealing with UI but the Universal theme is also kind of annoyingly space hungry, which is why I started to add compact variants. Though at some point it becomes more appropriate to ditch the Universal style entirely and roll a custom one. That was always the idea of going with QML, it forces logic and presentation to be split such that the two can be changed (mostly) independently of each other. So LLMs getting better at UI madness opens that direction.

WhiteMagic:
Oh other thing I remembered, the description of the option itself should be updated, to no longer mention that a restart is required.

SubliminalsTV:
Done — dropped "(requires restart)" from the dark-mode option's description now that it applies live, and generalized that comment to your wording since the signal isn't dark-mode-specific.

Re ditching the Universal style for a custom one — agreed that's exactly the payoff of QML's logic/presentation split; happy to help in that direction once you're further along with the compact variants.

qml/Main.qml
```diff
@@ -31,6 +31,16 @@ ApplicationWindow {
...
+
+    // Re-apply UI config (e.g. dark mode) whenever it changes, so toggling it
+    // in the options takes effect immediately instead of needing a restart.
+    Connections {
+        target: signal
+        function onConfigChanged() {
+            Style.isDarkMode = backend.useDarkMode
+        }
+    }
+
```
WhiteMagic: There is already a section in `Main.qml` where `Connections` are setup, including for the `signal` target. (around line 436) so this should go there.

_Edit: Somehow GitHub didn't post the comment yesterday :-/_

SubliminalsTV:
Moved the `onConfigChanged` handler into the existing `Connections { target: signal }` block — no standalone block now. (The `option.py` comment generalization from your earlier note is already in from the last push.)

WhiteMagic:
> Re ditching the Universal style for a custom one — agreed that's exactly the payoff of QML's logic/presentation split; happy to help in that direction once you're further along with the compact variants.

I'm unlikely going to add more custom compact variants, what is there fixes some of the main annoyances but replicating and condensing more Universal controls is to me more of a waste of time when it's all going to be replaced by an entirely new style anyhow. I'll have to write up a design doc for how that change might happen at some point.

PR #783, opened 2026-06-16
Make inactive device tab labels legible in dark mode

SubliminalsTV:
`JGTabButton` only styled the tab background, so inactive labels fell back to Universal's default dimming and read as disabled. Active tab now uses the full foreground colour, inactive tabs use the palette's `medColor` — legible but clearly unselected. Uses existing `Style` palette roles only.


WhiteMagic:
👍 looks good.

PR #794, opened 2026-07-08
Allow keyboard to Logical device "Button" mapping

emcodem:
Yea its a pure Vibe coded addition but it appeared relatively clean to me so i thought i open the PR...
If this makes sense or not depends on the usecase, mine is to reproduce issues of users using gremlin and vjoy in my own game - i don't have the same device the user has but using this functionality, i was able to reproduce their problem.

### DIFF
--- a/action_plugins/map_to_logical_device/__init__.py
+++ b/action_plugins/map_to_logical_device/__init__.py
@@ -248,17 +248,23 @@ class MapToLogicalDeviceData(AbstractActionData):
     ) -> None:
         super().__init__(behavior_type)
 
+        # LogicalDevice has no keyboard-type input, keyboard presses behave
+        # like a button press/release, so map them onto a logical button.
+        logical_type = behavior_type
+        if logical_type == InputType.Keyboard:
+            logical_type = InputType.JoystickButton
+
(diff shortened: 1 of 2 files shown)


WhiteMagic:
It would be useful to describe what this is actually meant to do. From what I can tell it's two things, fix a but and then add new functinoality. The bug is fixed somewhat correctly from a quick glance. The new functionality I don't see much purpose for as that can already be achieved via macros. Gremlin is strict with only letting devices control aspects that fit their physical behavior. Everything that is not a native / singular correct behavior shouldn't be a default.

WhiteMagic:
From what I can tell the bug fix fixes and issue but not in the right way, it re-invents what already exists while the second bit in device.py entirely misunderstand what goes on in that part of the code.

emcodem:
ok this was the firat and last time i try a vibe coded PR ;) 
as i dont need any of this anymore. i'd just close it, sorry for wastin your time buddy

WhiteMagic:
Wasn't really wasted as you pointed me to a bug, the fix was just a bit different due to non-obvious usages of what behavior etc. mean to Gremlin and its ui.

PR #801, opened 2026-08-01
Add handle editing, cursor readout and 0 to 100 scale to response curves

fredmayor88:
Adds numerical and visual control over cubic Bezier response curves.

Response curve editor with the new handle controls, cursor readout and 0 to 100 scale

## What's new

**Mirror handles** — per control point option keeping its two handles reflected through the point, so the curve leaves it at the same slope on both sides wherever the point is moved. Dragging one handle moves the other only while it is enabled.

**Numerical handle editing** — slope and length of the selected point's handles, previously reachable only by dragging.

**Cursor readout** — hovering the curve reports the input position, the resulting output, and the slope there.

**0 - 100 scale** — optional display scale for the numerical inputs and the readout. It converts on the way in and out and never touches stored data. Slopes are unaffected by it, since both axes scale equally.

## Backwards compatibility

Nothing existing changes.

- Every new option defaults to off. With them off the editor behaves exactly as it does today: handles stay independent, values read on -1 to 1.
- No existing control changed its range, step or precision.
- The runtime evaluation path is untouched.
- Profiles written before this change load identically, and re-saved untouched they serialize byte for byte the same. Handle symmetry is only written when at least one point uses it.
- Profiles written with handle symmetry load correctly in older releases. The flag lives inside `control-points` as `symmetric-handles` properties, and existing code reads `point` entries by name and ignores anything else.

`test/unit/test_action_response_curve.py` asserts each of these against a fixture profile in the pre-change format, including rebuilding the curve from the point entries alone the way an older release would.

## Also included

`ControlPoint.handleLeft` / `handleRight` returned `None` for the handle an end point does not have, which fails its QML type conversion and logs `Shiboken::Conversions::_pythonToCppCopy: Cannot copy-convert ... (NoneType) to C++` on every rebuild of the control point list. They now return an empty point, which is what the failed conversion produced anyway, so rendering is unchanged. Separable from the rest if you would rather take it on its own.

## Testing

`test/unit`, `test/integration/action_plugins` and `test/action_interaction` pass. `ruff check .` is clean.


fredmayor88:
Hey that was claude-written but I gave it a lot of specific instructions, to follow the codebase patterns, stick to the scope, ensure backwards compatibility etc. I also tried to add all of that in a way that wouldn't "scare" existing users. 
Adding this as I have a use case where I want to create a mapping curve for my simracing pedals and I need precise settings and to be able to know the exact slope of the curve. I hope it works.

WhiteMagic:
Generally speaking most of these additions are fine. The main part is that this is too mcuh at once, at work such an MR would simply be rejected for being too big a change and too many separate  bits rolled into one. LLMs have no issue with that but someone reviewing it will just waste their time.

So the 4 respectively 5 points should be individual MRs. The parts that I'm not certain of, i.e. if they serve a good enough purpose to be added are:
- 0 - 100 display, I understand where this comes from but it will make debugging and helping people an even bigger issue
- Handle positioning control, I'm uncertain that level of control is required, and if it is it would likely be in x, y coordinates 

Other things I already noticed by skimming over it:
- The code suffers from typical LLM over commenting. Comments that explain/justify are useless as permanent code comments, but would be useful when adding them into the MR so I can understand what you intended with that, there is a `comment-code` skill that makes the code more in line with how code is documented in here. The skill leans to underdocument so you have to add missing parts, which is desirable imo.
- The test also looks like the usual "llm goes and writes a test, resulting in long overly broad testing", which means it likely should be looked at and made more concise.
- "Mirror handles was called "Force smooth curves" in R13, which may or may not be a better name
- `hints.csv` that the LLM dropped stuff into is not used anywhere, it's a remenant of an old R13 "help system"

fredmayor88:
Thanks for taking a look! 
The things you mentioned you are not certain of are quire important for my use case. 
I see you have a certain shape in mind for this project, that's great of course. I am also seeing I am going to need a few more rather simracing specific features. 
So at this point I see this possible cooperation to contribute that upstream would require rather a lot of effort and likely result in not exactly what I have in mind from the final product perspective. Closing this and I will be happy with my own fork

WhiteMagic:
Even if all features made it, there would be a non trivial amount of work as I'm not merging in pure LLM generated code as it never is of good enough quality or built the way things truly should be in Gremlin. As a PoC fine, as a merged feature not.

fredmayor88:
if anybody made it here, I built this page to translate from (-1, 1) to (0, 100), have better control and control slopes:  

https://fredmayor88.github.io/gremlin-curve-converter

Issue #802, opened 2026-08-02
[R15] Vjoy Axis Condition Error / Treated as Button


djscott00:
**Description**
Two Problems when setting up a vJoy axis condition:

1) Selecting an axis from the input drop down will generate an exception of:
"Traceback (most recent call last):
   File "action_plugins\condition\condition.py", line 407, in _set_vjoy_input_type
    self._create_comparator(self._sates[0].input_type)
                            ^^^^^^^^^^^
 AttributeError: 'VJoyCondition' object has no attribute '_sates'. Did you mean: '_states'?
"
Root cause appears to be a typo here: https://github.com/WhiteMagic/JoystickGremlin/blob/c36dbe47c535d96bf10aeb9e52833d67132a5f85/action_plugins/condition/condition.py#L407

Correcting line 407 to "_states" instead of "_sates" resolves the exception.

2) After correcting the above, the condition options don't update to the controls appropriate for an axis; it's still showing the condition options for a button.

**Expected behaviour**
Exception should not be thrown here, and selecting a vjoy axis should instead display separate controls to define the relevant range of the axis to target.

**Screenshots**
<img width="1438" height="667" alt="Image" src="https://github.com/user-attachments/assets/f4f77012-1ac7-4573-ae7c-be08fb8cd64b" />

**Information**
- Gremlin version: R15

djscott00:
Looks like both issues may be resolved by changing the line to just:
self._create_comparator(input_type_tmp)

WhiteMagic:
Thanks for finding and then fixing the bug 👍.

PR #803, opened 2026-08-02
#802 - Vjoy Axis Condition Fix

djscott00:
See: https://github.com/WhiteMagic/JoystickGremlin/issues/802

Fix for vjoy axis condition exception, and that it would display the condition controls for a button instead of an axis range.

This change now will actually show the range controls for the vjoy axis, and appears to be working as expected from a quick test here.

image


WhiteMagic:
Oh nice that's like two bugs in one :-) Typo and the wrong logical type as well.

Issue #805, opened 2026-08-02
New Merge Axis Operations: Prefer Extreme and Prefer Center


djscott00:
(I'll likely attempt to implement this and submit a PR, but getting this in for tracking...)
 
- Describe the problem this feature request addresses.
There's some gaps with the available merge operations, in particular, that we may want to use to input axes that both rest at 0.0, and we want to output whichever value is further from the center (or closest to the center). The "maximum" and "minimum" operations partially cover this, but only for the respective +/- sides.

- What does the use case for this new feature look like?
For the user, just new drop down selection option in the UI.
The "Prefer Extreme" will use the higher absolute value for the result.
The "Prefer Center" will use the lower absolute value for the result.

- Does this feature extend an existing one or is it entirely new?
Extends on existing "merge_axis" action plugin.

WhiteMagic:
Implemented in #806 and compelted in #829 .

PR #806, opened 2026-08-02
805 new merge operations

djscott00:
See #805: https://github.com/WhiteMagic/JoystickGremlin/issues/805

Added 2 new merge operations for "Preferextreme" and "Prefercenter" (I'd like to display these options as better looking strings in the drop-down, but didn't manage that for now...)

Appears to be working as expected from some quick tests with each.

### DIFF
--- a/action_plugins/merge_axis/__init__.py
+++ b/action_plugins/merge_axis/__init__.py
@@ -56,6 +56,7 @@ class MergeOperation(Enum):
     Sum = 3
     Bidirectional = 4
     Preferextreme = 5
+    Prefercenter = 6
 
     @classmethod
     def to_string(cls, value: MergeOperation) -> str:
@@ -66,6 +67,7 @@ class MergeOperation(Enum):
             MergeOperation.Sum: "sum",
             MergeOperation.Bidirectional: "bidirectional",
             MergeOperation.Preferextreme: "preferextreme",
+            MergeOperation.Prefercenter: "prefercenter"
         }
 
         res = lookup.get(value, None)
@@ -81,7 +83,8 @@ class MergeOperation(Enum):
             "maximum": MergeOperation.Maximum,
             "sum": MergeOperation.Sum,
             "bidirectional": MergeOperation.Bidirectional,
-            "preferextreme": MergeOperation.Preferextreme
+            "preferextreme": MergeOperation.Preferextreme,
+            "prefercenter": MergeOperation.Prefercenter
         }
         res = lookup.get(value.lower(), None)
         if res is None:
@@ -147,9 +150,16 @@ class MergeAxisFunctor(AbstractFunctor):
     @staticmethod
     def _preferextreme(value1: float, value2: float) -> float:
         """Merges two axes into one:
-            - the axis furthest from center is used
+            - the axis furthest from center (0.0) is used
         """
         return value1 if abs(value1) > abs(value2) else value2
+
+    @staticmethod
+    def _prefercenter(value1: float, value2: float) -> float:
+        """Merges two axes into one:
+            - the axis closest to center (0.0) is used
+        """
+        return value1 if abs(value1) < abs(value2) else value2    
 
     actions = {
         MergeOperation.Average: _average,
@@ -158,6 +168,7 @@ class MergeAxisFunctor(AbstractFunctor):
         MergeOperation.Sum: _sum,
         MergeOperation.Bidirectional: _bidirectional,
         MergeOperation.Preferextreme: _preferextreme,
+        MergeOperation.Prefercenter: _prefercenter
     }
 
 
--- a/action_plugins/merge_axis/__init__.py
+++ b/action_plugins/merge_axis/__init__.py
@@ -55,6 +55,7 @@ class MergeOperation(Enum):
     Maximum = 2
     Sum = 3
     Bidirectional = 4
+    Preferextreme = 5
 
     @classmethod
     def to_string(cls, value: MergeOperation) -> str:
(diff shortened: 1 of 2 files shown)


action_plugins/merge_axis/__init__.py
```diff
@@ -141,12 +147,28 @@ def _bidirectional(value1: float, value2: float) -> float:
...
 
+    @staticmethod
+    def _preferextreme(value1: float, value2: float) -> float:
+        """Merges two axes into one:
+            - the axis furthest from center (0.0) is used
+        """
+        return value1 if abs(value1) > abs(value2) else value2
+
+    @staticmethod
+    def _prefercenter(value1: float, value2: float) -> float:
```
WhiteMagic, line 158: If we rename it to `_closest_to_center` it's obvious and the comment isn't needed anymore (the reason only bidirectional is documented is that it's not self-explanatory). So with the new name that is clear on its own.

action_plugins/merge_axis/__init__.py
```diff
@@ -141,12 +147,28 @@ def _bidirectional(value1: float, value2: float) -> float:
         """
         return (value2 - value1) / 2.0
 
+    @staticmethod
+    def _preferextreme(value1: float, value2: float) -> float:
```
WhiteMagic, line 151: I did some name searching as extrema makes sense in my head from a mathematical point of view but likely isn't obvious to people. The best name I could come up with was `_maximum_magnitude` which again is self explanatory as magnitude is well defined as `abs(value)`. Though not sure if that's a well enough understood term in general.

action_plugins/merge_axis/__init__.py
```diff
@@ -141,12 +147,28 @@ def _bidirectional(value1: float, value2: float) -> float:
...
+        return value1 if abs(value1) < abs(value2) else value2    
+
     actions = {
         MergeOperation.Average: _average,
         MergeOperation.Minimum: _minimum,
         MergeOperation.Maximum: _maximum,
         MergeOperation.Sum: _sum,
         MergeOperation.Bidirectional: _bidirectional,
+        MergeOperation.Preferextreme: _preferextreme,
+        MergeOperation.Prefercenter: _prefercenter
```
WhiteMagic, line 171: If we rename them, I would use CamelCase for them, i.e. `MaximumDeflection` and `ClosestToCenter`.

action_plugins/merge_axis/__init__.py
```diff
@@ -79,6 +83,8 @@ def to_enum(cls, value: str) -> MergeOperation:
             "maximum": MergeOperation.Maximum,
             "sum": MergeOperation.Sum,
             "bidirectional": MergeOperation.Bidirectional,
+            "preferextreme": MergeOperation.Preferextreme,
+            "prefercenter": MergeOperation.Prefercenter
```
WhiteMagic, line 87: For XML naming I usually end up going for `-` to separate words, so `closest-to-center`.

gh_page/interface/actions.md
```diff
@@ -212,6 +212,12 @@ The manner in which the axes values are combined is governed by the *Merge opera
 
     <dt>Sum</dt>
     <dd>Computes the sum of the two values.</dd>
+
+    <dt>Prefer Extreme</dt>
+    <dd>Computes the largest absolute value of the two values.</dd>
+
+    <dt>Prefer Center</dt>
+    <dd>Computes the smallest absolute value of the two values</dd>
```
WhiteMagic, line 220: Nice 👍

WhiteMagic:
Added some comments mainly around the names, to make them hopefully understandable without looking at the document (which people tend not to do). The one part that should be added is integrated this into the "action interaction tests". For the axis merge action https://github.com/WhiteMagic/JoystickGremlin/blob/develop/test/action_interaction/test_merge_axis.py this is actually quite straightforward as the test isn't very elegant (no parametrization etc). 

All you need to do is copy one of the existing ones and adjust some values:
- Use the `set_merge_mode` helper to configure the action for the particular mode
- Keep or modify the existing values and adjust the  expected result according to the merge action

What these tests do is to load a profile that has actions defined on logical inputs with only logical inputs being used as mapping targets, i.e. purely virtual, no vjoy or physical device. Then I mangled an existing QT testing framework to be able to listen to these logical device events and expose them to the tests in a somewhat not too painful manner and allow interacting with the inputs, i.e. press, wait, etc.

action_plugins/merge_axis/__init__.py
djscott00: These lines are for: ```to_enum(cls, value: str) -> MergeOperation```

I'm finding that if I have a enum value of ```ClosestToCenter```, the corresponding string value that that's used with this method will be ```"Closesttocenter"```, due to ```e.name.capitalize()``` from the ```operationList(self) -> LabelValueSelectionModel``` method.

So I'm inclined to say that I can't use ``` - ``` characters here in the set of expected strings, at least not without reworking how other parts of merge_axis operates, so that the input strings weren't strictly based on the literal enum value...?

WhiteMagic: Ugh oh yes, in that case I would actually suggest we modify how this works, as the `.capitalize()` happens to work for the single word operations we have but they're not really useful to present to a human. 

The `to_string()` is used to serialize it to XML and is then co-opted with the `.capitalize()` to work as the display representation as well. The `to_enum` assumes it can get the canonical representation back using `.lower()`. So if we introduce a third dictionary like `to_from_display(str) -> str` which maps from the canonical to the display representation and vice versa that should work and allow whatever string we want. Bit more involved, and if it was purely the XML value I wouldn't care too much. But this also affects what the user sees which is why in my view it's worth adding this layer.

WhiteMagic:
Finished adding the display layer here #829.

PR #812, opened 2026-08-19
Fix data file collection for the Kobold style

SubliminalsTV:
Freezing `feature/kobold-style` with PyInstaller currently fails outright:

```
ERROR: Unable to find 'qml' when adding binary and data files.
```

`datas` still collects `qml` and `theme`, and neither directory exists on this branch any more. Removing them gets the build to complete, but the result won't start: `style/qml` is added to the QML import path at runtime and `QQuickStyle.setStyle("Kobold")` needs it, yet `style` was never collected, so the frozen app has no style to load.

This replaces the two dead entries with `style`. Themes need nothing extra since they resolve from `:/themes` in `resources.py`.

Verified by building this branch and confirming `_internal/style/qml/Kobold` and `_internal/style/themes` are present; I've also been running the same change in my own build to exercise the UI.

Only the data file list is touched, no packaging behaviour beyond that.


WhiteMagic:
Good catch, it was on the merge list for the UI rework. The actual change was a bit more specific, as taking the full style folder would ship unnecessary aspects.

PR #824, opened 2026-09-13
Clarify the description on the axis merge buttons

ELFoglalt:
I spent a day trying to set up an axis merge, and ended up reading the source code to figure out how to do this, because my brain just couldn't understand the UI.

I kept thinking that the "Rec" button should record an axis that I move, and that it assigning the current axis is caused by a problem with my setup. In my head Joystick Gremlin kept thinking that the current axis is being actuated and picking it immediately as the result of the recording.

Clearly this is not what is happening, and the code does exactly what it should. But I think there is a mismatch between the expectation of a record button, and what it actually does in this instance. This could be improved making the button say something else, for example "Assign Current Axis" or just "Assign".

The doc screenshots don't look exactly like my setup, so I'm not sure if I should touch them.


WhiteMagic:
The UI got completely overhauled in the meantime so your changes can't be applied but I included parts of the description into how things are now shown on that action.

